PR #40, opened 2017-10-06
updating sandbox configs, this later needs to be compiled dynamically…

Harsimran1:
… depending on environment


abetterway2feel:
Please follow https://chris.beams.io/posts/git-commit/

src/config.js
```diff
@@ -1,13 +1,13 @@
...
 		keySize: 128,
 	},
 	signinState: 'pkyyf',
 	zkit: {
-		service_url: 'https://xfjx51bvoe.api.tresorit.io',
-		clientId: 'xfjx51bvoe_XyFJHpD7L5',
+		service_url: 'https://g6y0wg1tf6.api.tresorit.io',
+		clientId: 'g6y0wg1tf6_RGxRknA5dI',
 	},
 };
```
abetterway2feel: this won't break terra, right? Ad the sdk isn't dynamically injected?

Harsimran1: No it won't. It is not dynamically injected. It is built locally and put inside terra, if there is anything new in SDK. We can host the SDK somewhere to add it dynamically.

PR #44, opened 2017-10-17
add fhir validator

Harsimran1:


src/lib/fhirValidator.js
```diff
@@ -0,0 +1,86 @@
...
+						result.errors = result.errors.concat(validate.errors);
+					}
+				} else {
+					result.errors.push(new ValidationError('Not a valid resource type'));
+				}
+				return result;
+			})
+			.catch(() => {
+				result.errors.push(new ValidationError('Not a valid resource type'));
+				return result;
```
isalga: Why don't you better throw an error instead of pushing to results and return it?

src/services/FhirService.js
```diff
@@ -10,8 +12,14 @@ class FHIRService {
 	}
 
 	uploadFhirRecord(JsonFHIR, tags = []) {
-		tags = [...tags, ...(new Tags()).createTagsFromFHIR(JsonFHIR)];
-		return this.uploadRecordWithTags(JsonFHIR, tags);
+		return fhir.validate(JsonFHIR)
+			.then((result) => {
+				if (result.errors && result.errors.length > 0) {
+					return Promise.reject(result.errors);
```
isalga: Same, if you throw an error in validate then you will not need to handle this.

src/lib/fhirValidator.js
```diff
@@ -0,0 +1,86 @@
...
+				this.conformance = {
+					types: schema.types,
+					resources: schema.resources,
+				};
+				return this.conformance;
+			});
+	}
+
+	isValidResourceType(resourceType) {
+		if (resourceType.includes('.')) return new Promise((resolve, reject) => reject(false));
```
isalga: Would be better less redundant: return Promise.reject(new Error(false))

src/lib/fhirValidator.js
```diff
@@ -0,0 +1,86 @@
...
+		return this.isValidResourceType(resource.resourceType)
+			.then((res) => {
+				if (res === true) {
+					const validate = ajv.compile(this.conformance.resources[resource.resourceType]);
+					validate(resource);
+					if (validate.errors) {
+						result.errors = result.errors.concat(validate.errors);
+					}
+				} else {
+					result.errors.push(new ValidationError('Not a valid resource type'));
```
isalga: same, throw error will be more clear.

src/lib/fhirValidator.js
```diff
@@ -0,0 +1,86 @@
...
+			else if (typeof data === 'object') resource = data;
+			else {
+				result.errors.push(new ValidationError(
+					'Data provided must be a string or plain javascript object',
+				));
+				return Promise.reject(result);
+			}
+			result.resource = resource;
+		} catch (e) {
+			result.errors.push(new ValidationError('Invalid JSON', e));
```
isalga: meh

src/lib/fhirValidator.js
```diff
@@ -0,0 +1,86 @@
...
+		// Ensure that the data provided is a javascript object.
+		// If it is a string then attempt to parse it
+		try {
+			if (typeof data === 'string') resource = JSON.parse(data);
+			else if (typeof data === 'object') resource = data;
+			else {
+				result.errors.push(new ValidationError(
+					'Data provided must be a string or plain javascript object',
+				));
+				return Promise.reject(result);
```
isalga: throw new ValidationError() or return Promise.reject(new ValidationError()) instead of returning result

src/lib/fhirValidator.js
```diff
@@ -11,75 +11,64 @@ class Fhir {
...
-				result.errors.push(new ValidationError(
-					'Data provided must be a string or plain javascript object',
-				));
-				return Promise.reject(result);
+			if (resource && resource.resourceType) {
+				this.isValidResourceType(resource.resourceType)
+					.then(() => {
+						const validate = ajv
+							.compile(this.conformance.resources[resource.resourceType]);
+						validate(resource);
```
isalga: Please rename validate variable

src/lib/fhirValidator.js
```diff
@@ -0,0 +1,75 @@
...
+
+const ajv = new Ajv({ extendRefs: true });
+
+class Fhir {
+	constructor() {
+		this.getConformance();
+	}
+	getConformance() {
+		return new Promise((resolve, reject) => {
+			if (this.conformance !== undefined) {
```
Glumli: Isn't 
"if (this.conformance) "
enough?

src/lib/fhirValidator.js
```diff
@@ -0,0 +1,75 @@
...
+				resolve(this.conformance);
```
Glumli: return after resolving

src/lib/fhirValidator.js
```diff
@@ -0,0 +1,75 @@
...
+					};
+					resolve(this.conformance);
+				})
+				.catch(reject);
+		});
+	}
+
+	isValidResourceType(resourceType) {
+		return new Promise((resolve, reject) => {
+			if (resourceType.includes('.')) reject(false);
```
Glumli: Why do you check for this?

src/lib/fhirValidator.js
Harsimran1: yes 👍

PR #47, opened 2017-10-18
Implement deleteRecord/Document

Glumli:


src/HealthCloud.js
```diff
@@ -19,6 +19,8 @@ class HealthCloud {
 		this.searchRecords = fhirService.searchRecords.bind(fhirService);
 		this.uploadFhirRecord = fhirService.uploadFhirRecord.bind(fhirService);
 		this.downloadFhirRecord = fhirService.downloadFhirRecord.bind(fhirService);
+		this.deleteRecord = fhirService.deleteRecord.bind(fhirService);
+		this.deleteDocument = fhirService.deleteRecord.bind(fhirService);
```
Harsimran1: why are there 2 methods?

test/services/fhirServiceTest.js
```diff
@@ -161,6 +161,30 @@ describe('fhir service', () => {
 		});
 	});
 
+	it('deleteRecords succeeds when userId is passed', (done) => {
```
Harsimran1: deleteRecord*

src/HealthCloud.js
Glumli: Currently deleteRecord is the same as deleteDocument, but the methods might change in the future. Still I think we should expose both, as the user doesn't know that it's the same

PR #54, opened 2017-10-26
Add update user

isalga:


src/services/UserService.js
```diff
@@ -38,6 +38,36 @@ class UserService {
 				return this.user;
 			});
 	}
+
+	updateUser(params) {
+		return new Promise((resolve, reject) => {
+			const userId = this.getUserId();
+
+			if (!userId) {
```
Harsimran1: Error message can be more standardized, start with capital letters and come from class type in lib/errors. Something like NotLoggedInError and ValidationError.

PR #58, opened 2017-11-01
Implement Logout

Harsimran1:


src/services/ZeroKitAdapter.js
```diff
@@ -132,6 +132,17 @@ class ZeroKitAdapter {
...
 				.then(() => tresorId)));
 	}
+
+	logout() {
+		return this.zeroKit.then(zerokit => zerokit.logout())
+			.then((res) => {
+				sessionHandler.deleteCookie('HC_Id');
+				sessionHandler.deleteCookie('HC_Auth');
+				sessionHandler.deleteCookie('HC_User');
+				sessionHandler.deleteCookie('HC_Refresh');
```
isalga: should this not better be done in a function on sessionHandler called 'reset'?

PR #60, opened 2017-11-02
Fix SDK upload, update Fhir, call latest fhir schema,

Harsimran1:


src/services/DocumentService.js
```diff
@@ -34,6 +34,7 @@ class DocumentService {
 	}
 
 	uploadDocument(userId, files, metadata = {}, customTags = []) {
+		customTags = [...customTags, ...['type:document']];
```
Glumli: Putting 'type:document' in an array for using the spread syntax then seems unnecessary for me. Why didn't you just go for '[...customTags, 'type:document']'

src/services/DocumentService.js
```diff
@@ -68,7 +69,8 @@ class DocumentService {
 						this.createDocumentReference(info.id, files[index].title));
 				documentRecord.body.content.push(...newAtachements);
 				return this.fhirService
-					.updateFhirRecord(documentRecord.record_id, documentRecord.body);
+					.updateFhirRecord(documentRecord.record_id,
+						documentRecord.body, documentRecord.tags);
```
Glumli: If you go for a new line within the brackets I would prefer a new line for each argument. But don't need to change this.

src/services/FhirService.js
```diff
@@ -14,7 +14,16 @@ class FHIRService {
...
 
-		return this.uploadFhirRecord(jsonFHIR, tags, updateRequest);
+		return this.downloadFhirRecord(recordId)
+			.then((res) => {
+				Object.keys(jsonFHIR).forEach((key) => {
+					if (Array.isArray(jsonFHIR[key]) && jsonFHIR[key].length === 0) {
+						delete jsonFHIR[key];
+					}
+				});
+				const newFhirJson = Object.assign(res.body, jsonFHIR);
```
Glumli: I thought we would just upload the new fhirJson for now and later think about if it might be necessary to merge the jsons some other way. 
As it is now, there is no way to delete an attribute from sol.

src/services/DocumentService.js
```diff
@@ -83,6 +85,12 @@ class DocumentService {
 			});
 	}
 
+	updateDocumentMetadata(recordId, jsonFHIR, tags = []) {
+		// don't update the content as it is internally handled by SDK
+		if (jsonFHIR.content) delete jsonFHIR.content;
```
Glumli: Maybe we should throw an error here, to tell the user that he he updates the files wrong. 
But for now that's fine.

PR #67, opened 2017-11-16
Refactor get login and register form to promise

isalga:


src/services/ZeroKitAdapter.js
```diff
@@ -26,29 +26,33 @@ class ZeroKitAdapter {
...
+				zkitLoginNode.removeChild(zkitLoginNode.firstChild);
+			}
+			const zKitLoginObject =
+				this.zeroKit
+					.then(zeroKit => zeroKit.getLoginIframe(zkitLoginNode));
+
+			const submit = function (zKitLogin, event) {
+				event.preventDefault();
+				this.login(zKitLogin, document.getElementById(loginFormIds.hcUserLogin).value)
+					.then(resolve)
```
Glumli: Why don't you just return this.login()?

src/services/ZeroKitAdapter.js
```diff
@@ -82,24 +85,28 @@ class ZeroKitAdapter {
...
+			while (zkitRegisterNode.firstChild) {
+				zkitRegisterNode.removeChild(zkitRegisterNode.firstChild);
+			}
+			const zKitRegistrationObject =
+				this.zeroKit.then(zeroKit => zeroKit.getRegistrationIframe(zkitRegisterNode));
+
+			const submit = function (zKitRegistration, cb, event) {
+				event.preventDefault();
+				const userAlias = document.getElementById(registrationFormIds.hcUserRegister).value;
+				this.register(zKitRegistration, userAlias)
```
Glumli: Same as for login

src/services/ZeroKitAdapter.js
isalga: It does not resolve the promise that way

src/services/ZeroKitAdapter.js
isalga: same

PR #68, opened 2017-11-16
Remove registration autovalidation

isalga:


src/routes/userRoutes.js
```diff
@@ -6,10 +6,7 @@ const apiUrl = config.api;
 const userRoutes = {
 
 	initRegistration(hcUserAlias) {
-		const body = {
-			email: hcUserAlias,
-			user_data: { auto_validate: true },
-		};
+		const body = { email: hcUserAlias };
```
Harsimran1: i hope it doesn't affect the PUT /user request, since it is updates user_data. You can verify before merging.

isalga: Tested it stills works properly

PR #69, opened 2017-11-17
Fix delete File

Glumli:


src/services/DocumentService.js
```diff
@@ -98,7 +98,7 @@ class DocumentService {
 		return this.fhirService.downloadFhirRecord(documentId)
 			.then((record) => {
 				record.body.content = record.body.content.filter(documentReference =>
-					fileIds.indexOf(documentReference.attachment.url) === -1);
+					fileIds.indexOf(documentReference.attachment.id) === -1);
```
Harsimran1: would be good idea if it is covered by test case.

PR #70, opened 2017-11-21
Adds proper tagging

Glumli:


src/HealthCloud.js
```diff
@@ -12,9 +12,11 @@ class HealthCloud {
 		const documentService = new DocumentService({ zeroKitAdapter });
 		const fhirService = new FhirService({ zeroKitAdapter });
 		UserService.setZeroKitAdapter(zeroKitAdapter);
+
 		this.getLoginForm = zeroKitAdapter.getLoginForm.bind(zeroKitAdapter);
 		this.getRegistrationForm = zeroKitAdapter.getRegistrationForm.bind(zeroKitAdapter);
 		this.downloadDocument = documentService.downloadDocument.bind(documentService);
+		this.getAllDocuments = documentService.getAllDocuments.bind(documentService);
```
Harsimran1: Looks like the endpoint for this has changed to users/{userId}/records?queryparms .  I would suggest to name it as getDocuments because later on , it might be used to filter documents and we don't want to rename the function again.

src/services/UserService.js
```diff
@@ -42,16 +42,16 @@ class UserService {
...
 			this.resolveUser()
-				.then((res) => {
-					user.email = res.email;
-					user.id = res.id;
+				.then((resolvedUser) => {
+					user.email = resolvedUser.email;
+					user.id = resolvedUser.id;
 
-					if (!res.user_data.encrypted_data) {
+					if (!resolvedUser.user_data.encrypted_data) {
```
isalga: Merge conflict probably here. There was a contition: `!res.user_data && !res.user_data.encrypted_data) ` that is important

PR #72, opened 2017-11-22
Fix refresh tocken call

isalga:


src/lib/hcRequest.js
```diff
@@ -23,7 +23,8 @@ const sendRefreshToken = () => {
 
 const isAuthorisedPath = path => ['documents', 'records'].some(el => path.includes(el));
 
-const isExpired = error => error.status === 401 && error.message.includes('expired');
+const isExpired = error =>
+	error.status === 401 && error.res && error.res.header['www-authenticate'].includes('expired');
```
Glumli: Can' this throw an exception when headers is undefined or there is no "www-authenticate" header?

PR #75, opened 2017-11-27
Add HCDocument

Glumli:


src/services/DocumentService.js
```diff
@@ -30,97 +28,87 @@ class DocumentService {
...
+				return this.updateDocument(userId, hcDocument);
+			});
 	}
 
-	addFilesToRecord(documentRecord, userId, files) {
-		let uploadInformation;
+	updateDocument(userId, hcDocument) {
+		let newAttachments = [];
+		let oldAttachments = [];
+		hcDocument.attachments.forEach((attachment) => {
```
Harsimran1: how are we making sure that the user who is trying to update metadata doesn't just delete his attachments by passing an empty list by mistake?

src/services/DocumentService.js
```diff
@@ -30,97 +28,87 @@ class DocumentService {
...
-	}
-
-	addFilesToDocument(userId, documentId, files) {
-		return this.fhirService.downloadFhirRecord(documentId)
-			.then(record => this.addFilesToRecord(record, userId, files));
+	uploadDocument(userId, hcDocument) {
+		return this.fhirService.createFhirRecord(hcDocument.toFhirObject())
+			.then((record) => {
+				hcDocument.id = record.record_id;
+				return this.updateDocument(userId, hcDocument);
```
Harsimran1: Looks like two records(1 with old version) will be created in uploadDocument when it should be only 1.

src/services/DocumentService.js
```diff
@@ -30,97 +28,87 @@ class DocumentService {
...
 
 	getDocuments() {
 		return this.fhirService.searchRecords({
 			tags: [taggingUtils.buildTag('resourceType', 'documentReference')],
-		});
+		})
+			.then(records =>
+				records.map((record) => {
+					let hcDocument = HCDocument.fromFhirObject(record.body);
+					hcDocument.id = record.record_id;
```
Harsimran1: this can also happen inside fromFhirObject method

test/lib/models/documentTest.js
```diff
@@ -0,0 +1,49 @@
...
+	it('constructor creates proper document', () => {
+		let hcDocument = new HCDocument(documentParameters);
+		expect(hcDocument.attachments.length).to.equal(2);
+		expect(hcDocument.type).to.equal(type);
+		expect(hcDocument.creationDate).to.equal(date);
+		expect(hcDocument.author).to.equal(author);
+		expect(hcDocument.title).to.equal(title);
+	});
+
+	it.skip('converting a document to fhir and back doesn\'t'
```
Harsimran1: need to remove skip

docs/sdk.md
```diff
@@ -234,34 +220,66 @@ In the case of any error in uploading and downloading documents, the format is:
...
+Therefore you need to remove the corresponding attachments from the attachments of the hcDocument.
+It is important to call ``updateDocument('user_id', hcDocument)`` afterwards to sync the object with the GesundheitsCloud.
 
 ```javascript
-    HC.deleteFilesFromDocument(userId, documentId, fileIds)
-        .then((response) => {
-           // use document record which contains document id, status etc.
+    hcDocument.attachments.pop(0);
+
+    HC.updateDocument(userId, hcDocument)
```
isalga: same.

docs/sdk.md
```diff
@@ -234,34 +220,66 @@ In the case of any error in uploading and downloading documents, the format is:
...
-        .then((response) => {
-            // use document record which contains document id, status etc.
-        })
-        .catch((error) => {
-            // error contains the status and error message
-        });
+    let hcAttachment = new HC.models.Attachment({ file });
+    hcDocument.attachments.push(hcAttachment);
+
+    HC.updateDocument(userId, hcDocument)
```
isalga: Make sure that we update Terra when this is deployed.

src/HealthCloud.js
```diff
@@ -31,6 +30,11 @@ class HealthCloud {
 		this.updateFhirRecord = fhirService.updateFhirRecord.bind(fhirService);
 		this.deleteRecord = fhirService.deleteRecord.bind(fhirService);
 		this.logout = zeroKitAdapter.logout.bind(zeroKitAdapter);
+
+		this.models = {
+			HCDocument,
```
isalga: Is this a FHIR object model generator?

docs/sdk.md
```diff
@@ -138,6 +140,38 @@ You can edit user data from the currently logged in user.
  .catch((error) => {});
 ```
 
+### Documents
+
+#### Document
+All operations on Documents are based on the hcDocument.
+For creating a Document you can create it with those optional parameters:
```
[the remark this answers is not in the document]
Glumli: Actualy they are all optional at the moment. But maybe we should make files a required parameter.

docs/sdk.md
```diff
@@ -146,25 +180,16 @@ Only the owner of the documents and those who have explicit permission from the
 
 To upload the document:
 ```javascript
- let files = [new File()];
+let hcDocument = new HC.models.Document();
```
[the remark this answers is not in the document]
Glumli: We cann add a file in this example.
" Can we just not mention what should be the type of hcDocument. Why do they have to call the extra model creation method if it is easy to create?"
I don't get what you want to tell me :/ Can you please explain further?

src/services/DocumentService.js
Glumli: between line 46 and 50? Could you please tell me where? I don't see it

src/services/DocumentService.js
Glumli: Why should the user have a wrong document? I think we don't have to make it sure, but we can check wether attachments is empty, which we don't allow

test/lib/models/documentTest.js
Glumli: I know,  but it fails ;)

src/services/DocumentService.js
Glumli: setting the id? I thought about as well, but the Id is not part of the fhir record

docs/sdk.md
Glumli: 👍  will be updating Terra today

docs/sdk.md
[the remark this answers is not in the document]
Glumli: Not Optional in FHIR

Glumli:
@isalga could you please "unblock" it? Maybe the next time it's enough to talk to me and trust me? (I can merge anyways) ;)

Glumli:
@isalga changed the settings so you really blocked me now

src/lib/models/HCAttachment.js
```diff
@@ -6,6 +8,12 @@ class HCAttachment {
 		creationDate = file.lastModifiedDate,
```
isalga: Please remove this default values in a destructured arguments. If no argument is supplied code will crash. `TypeError: Cannot destructure property `file` of 'undefined' or 'null'`

PR #77, opened 2017-11-28
H/validate email

Harsimran1:


src/lib/validationUtils.js
```diff
@@ -0,0 +1,12 @@
+const emailRegex = /\S+@\S+\.\S+/;
+
+
+const validationUtils = {
```
Glumli: Maybe rather call it emailUtils? At least not validationUtils, as we also got the fhirValidator that is not in the validation Utils?

Harsimran1: we might not want to add different file for each validation that comes later on?

Glumli: but we might also not want to have some validations happening in the validation utils and others is an extra validator. But I don't mind. Keep it like this for now.
BUT:  Maybe the regex should be at least a little bit more complex. 
Did you think about just setting the type of the email-field to email?

Glumli: https://stackoverflow.com/questions/2049502/what-characters-are-allowed-in-an-email-address
https://stackoverflow.com/questions/46155/how-to-validate-email-address-in-javascript
Rather allow invalid emails than not allowing valid emails. After reading a little in the second link I would even allow your '/\S+@\S+\.\S+/'
Sorry

PR #78, opened 2017-11-29
Use Capella GET endpoint to fetch user details

astridej:
Currently, the SDK uses the /resolve endpoint on Capella to fetch
details for the current user such as the tresor ID or the tag encryption
key. This endpoint is only meant for translating between e-mails and
Gesundheitscloud/Zerokit user IDs and will shortly have its responses
changed to reflect this. Instead, the GET endpoint for the specific
user should be used for such requests. Note this requires the user to be
logged in.


src/services/UserService.js
```diff
@@ -66,12 +66,20 @@ class UserService {
 	}
 
 	resolveUser() {
+		const userId = this.getUserId();
```
Glumli: Why did you put it here and not after the if block?

astridej: Whoops! You're right, that should go after the if block - shall fix.

src/services/UserService.js
```diff
@@ -71,7 +71,15 @@ class UserService {
 				resolve(this.user);
 			});
 		}
-		return userRoutes.resolveUserId(this.getUserAlias())
+
+		const userId = this.getUserId();
+		if (!userId) {
+			return new Promise((resolve, reject) => {
+				reject(new LoginError(NOT_LOGGED_IN));
```
isalga: return Promise.reject(new LoginError(NOT_LOGGED_IN))

src/routes/userRoutes.js
```diff
@@ -26,6 +26,10 @@ const userRoutes = {
 		return hcRequest('POST', `${apiUrl}/users/resolve`, body);
 	},
 
+	getUserDetails(userId) {
+		return hcRequest('GET', `${apiUrl}/users/${userId}`);
```
isalga: How differs the response of `getUserDetails` from `resolveUserId`? If so I would prefer this to be merged after demo, because maybe terra needs to adapt.

astridej: Currently, the responses are identical.

PR #79, opened 2017-12-07
Unexposing record methods

Glumli:


src/lib/models/HCAttachment.js
```diff
@@ -0,0 +1,34 @@
...
+	constructor({
+		file,
+		title = file.name,
+		type = file.type,
+		creationDate = file.lastModifiedDate,
+	}) {
+		this.file = file;
+		this.title = title;
+		this.type = type;
+		this.creationDate = creationDate;
```
isalga: Why don't you validate the paramers? If not it will leat to invalid fhir objects

src/lib/models/HCAttachment.js
```diff
@@ -0,0 +1,34 @@
+class HCAttachment {
+	constructor({
+		file,
+		title = file.name,
```
isalga: What if file does not contain a name? or if it is not an object, or null (it would crash)

src/lib/models/HCDocument.js
```diff
@@ -0,0 +1,19 @@
...
+
+class HCDocument {
+	constructor({
+		files = [],
+		type = 'Document',
+		title,
+		creationDate = new Date(),
+		author,
+	} = {}) {
+		this.attachments = files.map(file => new HCAttachment({ file }));
```
isalga: Same as HCAttachement. Please validate parameters. I think this modeling layer would be the perfect spot to use types.

Glumli: Simran didn't want to use types in the sdk. But you're right. Wait a second :)

src/services/DocumentService.js
```diff
@@ -30,97 +30,94 @@ class DocumentService {
...
-	getDocuments() {
-		return this.fhirService.searchRecords({
-			tags: [taggingUtils.buildTag('resourceType', 'documentReference')],
-		});
+	getDocuments(params = {}) {
+		params.tags = [taggingUtils.buildTag('resourceType', 'documentReference')];
+		return this.fhirService.searchRecords(params)
+			.then(records =>
+				records.map((record) => {
+					const hcDocument = hcDocumentUtils.fromFhirObject(record.body);
```
isalga: Same, validate record.body

src/services/DocumentService.js
```diff
@@ -10,17 +13,14 @@ class DocumentService {
 	}
 
 	downloadDocument(userId, documentId) {
-		let documentRecord;
+		let hcDocument;
 		return this.fhirService.downloadFhirRecord(documentId)
 			.then((record) => {
-				documentRecord = record;
+				hcDocument = hcDocumentUtils.fromFhirObject(record.body);
```
isalga: Can you not validate if record.body is correct fhir before passing it to `fromFhirObject`?

src/lib/models/HCDocument.js
isalga: Yes but the argument of Simran was that types add extra lines of code, but this can be removed with https://github.com/flowtype/flow-remove-types

isalga: Anyway to use types or not is your decition, but if at least validate the functions inputs or otherwise the SDK will be fragile

Glumli: I think her argument was about the extra dependencie. But I'm really pro types.

src/services/DocumentService.js
Glumli: I wanted to trust our API :(

isalga: Trust no one @Glumli , trust no one... not even yourself.

Glumli: :(

Glumli: It is indirectly checked now in the constructor of HCDocument

PR #91, opened 2017-12-19
Prepare Cross User transactions

Glumli:


test/services/fhirServiceTest.js
```diff
@@ -10,253 +10,180 @@ import documentRoutes from '../../src/routes/documentRoutes';
...
 	it('searchRecords succeeds', (done) => {
-		const fhirResponseList = [rawFhirResponse];
-		const searchRecordsStub = sinon.stub(documentRoutes, 'searchRecords')
-			.returnsPromise().resolves(fhirResponseList);
-
+		const secondUserId = 'userId2';
+		const secondTag = 'tag2';
 		const params = {
-			user_ids: ['user1', 'user2'],
+			user_ids: [testVariables.userId, secondUserId],
```
Harsimran1: this looks inconsistent, you can add all the variables in the testVariable

test/services/fhirServiceTest.js
```diff
@@ -10,253 +10,180 @@ import documentRoutes from '../../src/routes/documentRoutes';
...
-
-		fhirService.downloadFhirRecord('fakeRecordId').then((res) => {
-			expect(res.body).to.deep.equal(fhirObject);
-			expect(downloadFhirStub).to.be.calledOnce;
-			expect(downloadFhirStub).to.be.calledWith('fakeUserId');
-			expect(userServiceResolveUserStub).to.be.calledOnce;
-			downloadFhirStub.restore();
-			done();
-		});
+	it('downloadFhirRecord - Happy Path', (done) => {
```
Harsimran1: do we have tests for sad paths?

Glumli: We never had tests for anything else then the Happy path for downloading.

test/services/fhirServiceTest.js
Glumli: I didn't want to put it in there, as it is only user in one test. But I can put it in there

src/services/UserService.js
```diff
@@ -10,104 +10,88 @@ class UserService {
...
 	}
 
-	getUser() {
+	isCurrentUser(userId) {
+		const currentUser = this.getCurrentUser();
+		if (!currentUser || currentUser.id !== userId) return false;
+		return true;
+	}
+
+	getInternalUser(userId, partial = false) {
```
Harsimran1: This name sounds weird to me. How about getUserById?

src/services/FhirService.js
```diff
@@ -61,14 +61,13 @@ class FHIRService {
 	}
 
 
-	downloadFhirRecord(recordId) {
-		return documentRoutes.downloadRecord(UserService.getUserId(), recordId)
+	downloadFhirRecord(ownerId, recordId) {
```
Harsimran1: All these methods will be breaking terra, need to change terra and documentation too before deployment.

Glumli: I know

src/services/UserService.js
Glumli: We somehow need to make a difference between the get User that is used internatlly and the one externally
getUserByIf would be fine for the external one

src/services/FhirService.js
Glumli: I actually forgot the documentation again, but I'm on the update of Terra

PR #93, opened 2017-12-20
Add register form styles

isalga:


Glumli:
But could you maybe wait with merging until my PR went through?

PR #94, opened 2017-12-21
adds client tag

Glumli:


Glumli:
WIP

src/services/FhirService.js
```diff
@@ -28,7 +28,10 @@ class FHIRService {
 	uploadFhirRecord(ownerId, fhirObject, uploadRequest) {
 		return fhirValidator.validate(fhirObject)
 			.then(() => {
-				const tags = taggingUtils.generateTagsFromFhirObject(fhirObject);
+				const tags = [
+					...taggingUtils.generateTagsFromFhirObject(fhirObject),
+					taggingUtils.buildTag('client', this.zeroKitAdapter.authService.clientId),
```
Harsimran1: i feel all the logic for building tags should be in taggingUtils

Glumli: The thing about this tag is, that it has nothing to do with the fhirRecord. Therefore if the taggingUtils would tag it theirself, they would then then have a dependency to the auth service.
I see the clientId beeing a part of the record in the future and therefore didn't want to make bigger changes than necessary now.

PR #98, opened 2017-12-22
Js/reduce calls on initial login

Glumli:


src/services/ZeroKitAdapter.js
```diff
@@ -64,11 +64,6 @@ class ZeroKitAdapter {
 	}
 
 	login(zKitLoginObject, hcUserAlias) {
-		if (!validationUtils.validateEmail(hcUserAlias)) {
```
Glumli: Iris and me have the opinion that we should do this check on login, but only on registration

test/services/zerokitAdapterTest.js
```diff
@@ -151,8 +145,10 @@ describe('zerokitAdapter', () => {
 				expect(zKitLoginObject.login).to.be.calledOnce;
 				expect(zkit_sdk.createTresor).to.be.calledOnce;
 				expect(addTagEncryptionKeyStub).to.be.calledOnce;
+				expect(UserService.user.tresorId).to.equal('fakeTresorId');
 				done();
-			});
+			})
+			.catch(console.log);
```
isalga: remove console.log please

src/services/ZeroKitAdapter.js
Harsimran1: why do we want to make any requests to backend if user entered invalid email (maybe by mistake)?

PR #100, opened 2018-01-04
Makes clientId search parameter

Glumli:


docs/sdk.md
```diff
@@ -319,6 +319,7 @@ By calling ``getDocuments`` all documents of the user logged in can be received.
 ```
 For filtering those Documents you can add an object of parameters to the call. Possible parameters are:
 ```
+ client_id (optional, string) ... clientId from which the records where uploaded.
```
Harsimran1: I would have liked it to be in tags though, adds code complexity and too much specialisation in the long run if we keep doing this.

Glumli: As we currently don't expose tags I can't to it currently. As soon as tags I would love to put it there

PR #101, opened 2018-01-04
Js/fixes request bugs

Glumli:


src/routes/userRoutes.js
```diff
@@ -31,7 +31,7 @@ const userRoutes = {
 	},
 
 	getGrantedPermissions(userId, granteeId) {
-		return hcRequest('GET', `${apiUrl}/users/${userId}/permissions?grantee_id=${granteeId}`);
+		return hcRequest('GET', `${apiUrl}/users/${userId}/permissions?grantee_id=${granteeId}`, { authorize: true });
```
Harsimran1: maybe add a test case where it fails if not authorised.

Glumli: Thought about it, but seemed like quite a unhelpful test to me. Maybe I'll at least add it to all authorized requests

Glumli: Added tests

PR #109, opened 2018-01-10
Switching chrome to headless

ibihim:
In case it does not break anything, it would make it more pleasant to run the tests (no chrome browser poping up):

https://developers.google.com/web/updates/2017/06/headless-karma-mocha-chai
>Headless Chrome is a way to run the Chrome browser in a headless environment without the full browser UI. One of the benefits of using Headless Chrome (as opposed to testing directly in Node) is that your JavaScript tests will be executed in the same environment as users of your site. Headless Chrome gives you a real browser context without the memory overhead of running a full version of Chrome.


ibihim:
@Glumli @Harsimran1 How could the coverage go down? I just squashed my commits and changed the commit message to be more appropriate 😄

ibihim:
test this please

PR #111, opened 2018-01-12
use localstorage except of cookies

Glumli:


src/lib/sessionHandler/web.js
```diff
@@ -1,34 +1,23 @@
 const sessionHandler = {
+	localStorage: window.localStorage,
```
ibihim: hmpf, this does not look sexy... I mean I am not still up to date with the JS SDK and I mostly worked on with NodeJS (and not knowing much about babeljs / webpack magic), but usually we pasted in the window object on JS lib initialization and guarded it against stuff like being overwritten or undefined. Grabbing it like that feels not... cool. NullPointerException? 😄 

```JavaScript
const global = (function () {
  return this;
})();
```

But I think someone with more experience in the frontend could support or block this thought.

src/lib/sessionHandler/web.js
```diff
@@ -1,34 +1,23 @@
...
 
 	get(key) {
-		return this.getCookie(key);
+		return localStorage.getItem(key);
```
ibihim: What happens if there is no localStorage? E.g. IE8? 😝 Do we need stuff like that (https://developer.mozilla.org/en-US/docs/Web/API/Storage/LocalStorage) or is BabelJS securing us?

s-urbaniak: hmm, according to https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage IE8 supports local storage, no?

Glumli: Just read this as well.

src/lib/sessionHandler/web.js
s-urbaniak: I am literally learning, but here's a reference point on how to auto-detect the presence of the local storage object: https://github.com/STRML/react-localstorage/blob/0afa2e1/react-localstorage.js#L3.

src/lib/sessionHandler/web.js
ibihim: Wh00t. Then I might be wrong... or no... MDN is wrong!

> Compatibility
> Storage objects are a recent addition to the standard. As such they may not be present in all browsers

A line above:

> Was added in Firefox 3.5

Sorry my fault!

isalga: Actually percentage of users that has no access to local storage is 3.08%(3.06% of them from opera mini). So in that case we should implement a in-memory local storage. Anyway there is no reason to not to prevent a `null pointer exception`

font: caniuse.com

luisfarzati: +1 to in-memory fallback, there's a [nice polyfill on MDN](https://developer.mozilla.org/en-US/docs/Web/API/Storage/LocalStorage#Compatibility) that fallbacks to cookies but could easily be changed to use a local map.

Still, being Opera Mini the only non-supported browser...

test/lib/sessionHandler/webTest.js
```diff
@@ -15,30 +20,68 @@ const expect = chai.expect;
...
-        document.cookie = 'SQLiteManager_currentLangue=2';
-        document.cookie = ' HC_User=fakeUserId1,fakeUserAlias1';
-    });
+    let sessionHandler;
 
-    it('get sessionHandler succeeds', () => {
-        const userCookie = sessionHandler.get('HC_User');
-        expect(userCookie).to.equal('fakeUserId1,fakeUserAlias1');
-    });
+    describe('localStorage aviable', () => {
```
ibihim: typo 😄

test/lib/sessionHandler/webTest.js
```diff
@@ -15,30 +20,68 @@ const expect = chai.expect;
...
+        beforeEach(() => {
+            window.localStorage.clear();
+            sessionHandler = proxyquire('../../../src/lib/sessionHandler/web', {
+                '../environmentUtils': { default: { getLocalStorage: sinon.stub().returns(window.localStorage) } },
```
ibihim: Is it possible to get rid of proxyquire? In Simrans review, she removed a couple of those.

test/lib/sessionHandler/webTest.js
```diff
@@ -15,30 +20,68 @@ const expect = chai.expect;
...
+            }).default;
+        });
 
-    it('set sessionHandler succeeds', () => {
-        sessionHandler.set('HC_User', 'fakeUserId,fakeUserAlias');
-        expect(document.cookie).to.include('HC_User=fakeUserId,fakeUserAlias');
-    });
+        it('get succeeds', () => {
```
ibihim: it('should be possible to get values') is a better phrasing. it reads like 'it should be possible to get values.'

ibihim: this is true for all tests.

test/lib/sessionHandler/webTest.js
ibihim: this is for all proxyquires

src/lib/environmentUtils.js
```diff
@@ -0,0 +1,7 @@
+const environmentUtils = {
+    getLocalStorage() {
+        return window && window.localStorage ? window.localStorage : undefined;
```
ibihim: at some point it is worth to think to require lodash.get or so. but not a change request.

test/lib/sessionHandler/webTest.js
Glumli: I hate proxyquire. I think she moved them to the beforeEach block, because I asked her to do it ;)

test/lib/sessionHandler/webTest.js
Glumli: Great... used to it like this once... but getting old

Glumli: Great... used to it like this once... but getting old

PR #114, opened 2018-01-16
Fixing login

Glumli:


src/services/AuthService.js
```diff
@@ -43,13 +43,13 @@ class Auth {
...
             });
 
-            iframe.src = String.raw`${config.api}/auth/auth?                
-                client_id=${encodeURIComponent(this.clientId)}&
-                redirect_uri=${encodeURIComponent(window.location.origin + window.location.pathname)}&
-                response_type=code&
-                scope=everything&
-                state=${this.signInState}#
-                ${sessionHandler.get('HC_Id')}`;
+            iframe.src = `${config.api}/auth/auth?` +
```
Harsimran1: can you add a test case that the right format of url is going?

Glumli: Thought about it ;)

test/services/authServiceTest.js
```diff
@@ -92,13 +93,16 @@ describe('AuthService', () => {
...
-        authService.idpLogin().then((res) => {
-            expect(res.access_token).to.equal('fake_access_token');
-            expect(res.refresh_token).to.equal('fake_refresh_token');
-            expect(handleIframeStub).to.be.calledTwice;
-            expect(getAccessTokenFromCodeStub).to.be.calledOnce;
-            done();
-        });
+        authService.idpLogin()
+            .then((res) => {
+                expect(handleIframeStub.getCall(0).args[0].src.toLowerCase()).to.equal(`${config.api}/login?client_id=${testVariables.zeroKitId}&reto=http%3A%2F%2Flocalhost%3A9876%2Fcontext.html`.toLowerCase());
```
Harsimran1: where does this http://localhost:9876 come from? Can we replace it with some reference ? Not a change request.

PR #116, opened 2018-01-17
H/count records

Harsimran1:
Changes in PR: 

1. Use search endpoint GET /user/{userId}/records?queryParams instead of /records?queryParams. So change in the exposed method ```getDocument(params)``` to ```getDocuments('userId', 'params')``` and return type from ```[]``` to ```{
totalCount: 100,
records: []
}```

2. Add new method to fetch counts of the document.
 ```getDocumentsCount('userId', 'params')``` which returns ```{
totalCount: 100
}``` which uses HEAD /user/{userId}/records?queryParams

 Refer https://ares.epic.hpi.uni-potsdam.de/epicjira/jira/browse/HC-830


src/lib/hcRequest.js
```diff
@@ -43,7 +44,12 @@ const hcRequest = (type, path, {
             .query(query)
             .responseType(responseType)
             .send(body)
-            .then(res => res.body || res.text)
+            .then((res) => {
+                if (includeResponseHeaders) {
+                    return ({ body: res.body, headers: res.headers });
+                }
+                return res.body || res.text;
```
Harsimran1: I am not sure about this peace of code. Return types of the method seems to be different. Still looking for better approaches. One option is to return just `res` here and resolve whatever is needed in respective routes, but that doesn't look elegant either. Any suggestions are welcome.

test/services/fhirServiceTest.js
```diff
@@ -176,12 +204,13 @@ describe('services/fhirService', () => {
             tags: `${testVariables.encryptedTag}`,
         };
 
-        fhirService.searchRecords(params)
+        fhirService.searchRecords(testVariables.userId, params)
```
Harsimran1: writing more test cases

src/lib/hcRequest.js
luisfarzati: Hmm if the content type varies then I would add it in the response, so the consumer may use it... something like:

```
return { 
  type: res.headers['Content-Type'],
  body: res.body || res.text
}
```

or if we are sure it's always going to be either JSON or plain text, then maybe

```
return {
  type: res.body ? 'json' : 'text',
  body: res.body || res.text
}
```

src/services/DocumentService.js
```diff
@@ -109,15 +109,23 @@ class DocumentService {
...
+                        hcDocument.id = record.record_id;
+                        return hcDocument;
+                    });
+                }
+                return result;
+            });
+    }
+
+    getDocumentsCount(ownerId, params = {}) {
+        return this.getDocuments(ownerId, params, true);
```
isalga: Why does it not call directly to `documentRoutes.getRecordsCount()`? Now this count call depends on a parameter in `getDocuments` and `fhirService.searchRecords`. Which can potentially break in future tasks/refactors and makes the code a bit spaghetti 🍝 .

Also `getDocuments()` is exposing a way of just getting the document count. Is this intentional? It seems for me that this will allow SDK user to develop in a hacky way + not sure if this is a feature someone would expect when you also have a `getDocumentsCount` exposed.

Harsimran1:
I am adding the jira task id in the description, since we already have discussion in this pull request and it will be lost on renaming the branch.

src/services/DocumentService.js
Harsimran1: I added it here because most of the code will be repeated. But I agree with spaghetti code, makes sense to have it separately. Will make the change.

src/lib/hcRequest.js
Harsimran1: actually the consumers i.e our routes are just able to use the response as it is, whether it comes in `body` or `text`. But in case of `includeResponseHeader=true` condition, the consumer has to know that it has to do `result.body` https://github.com/gesundheitscloud/hc-sdk-js/blob/f55fe744b6b82933706b61af026e829a90a65933/src/routes/documentRoutes.js#L25-L28
Normally it is always https://github.com/gesundheitscloud/hc-sdk-js/blob/f55fe744b6b82933706b61af026e829a90a65933/src/routes/documentRoutes.js#L35-L37

luisfarzati: Ah ok, yeap that makes sense.

docs/sdk.md
```diff
@@ -325,6 +326,32 @@ For filtering those Documents you can add an object of parameters to the call. P
...
+```
+You can also get the total count based on filters as discussed above.
+The response will be an object with the totalCount parameter.
+
+```javascript
+   {
+        totalCount: 10
+   }
+```
+
```
ibihim: 👍

src/lib/hcRequest.js
```diff
@@ -30,6 +30,7 @@ const hcRequest = (type, path, {
     headers = {},
     responseType = '',
     authorize = false,
+    includeResponseHeaders = false,
```
ibihim: Wow. So many attributes. This is something we should change at some point.

src/services/FhirService.js
```diff
@@ -100,36 +89,41 @@ class FHIRService {
...
+                    return documentRoutes.getRecordsCount(ownerId, queryParams);
+                }
+                return documentRoutes.searchRecords(ownerId, queryParams);
+            })
+            .then((searchResult) => {
+                totalCount = searchResult.totalCount;
+                return searchResult.records ? Promise.all(searchResult.records
+                    .map(result => this.decryptRecordAndTags(result, user.tek))) : undefined;
+            })
+            .then(results => (results ? { totalCount, records: results } : { totalCount }));
```
ibihim: Same here.

test/routes/documentRoutesTest.js
```diff
@@ -114,6 +105,33 @@ describe('documentRoutes', () => {
                 `${config.api}/users/fakeUserId/records/fakeRecordId/status/Active`);
             done();
         });
+    });
+
+    it('searchRecord passes', (done) => {
```
ibihim: We had a story for test case naming, right? 😄

test/services/documentServiceTest.js
```diff
@@ -229,6 +234,59 @@ describe('DocumentService', () => {
         });
     });
 
+
+    describe('getDocuments', () => {
+        it('should pass', (done) => {
```
ibihim: beautiful

test/services/fhirServiceTest.js
```diff
@@ -29,7 +29,7 @@ const expect = chai.expect;
 describe('services/fhirService', () => {
     let zeroKitAdapter;
     let fhirService;
-    let authService = { clientId: testVariables.clientId };
+    const authService = { clientId: testVariables.clientId };
```
ibihim: I did the exact same thing too in my PR 😄

test/services/fhirServiceTest.js
```diff
@@ -147,20 +153,42 @@ describe('services/fhirService', () => {
...
-                expect(searchRecordsStub).to.be.calledWith(expectedParamsForRoute);
+                expect(searchRecordsStub).to.be.calledWith(testVariables.userId,
+                    expectedParamsForRoute);
+                expect(getInternalUserStub).to.be.calledOnce;
+                done();
+            })
+            .catch(done);
+    });
+
+    it('searchRecords - returns only count when one of parmas is countOnly', (done) => {
```
ibihim: parmas? spelling?

test/lib/hcRequestTest.js
```diff
@@ -47,7 +47,7 @@ describe('hcRequest', () => {
         }).default;
     });
 
-    it('hcRequest passes when api sends successful response', (done) => {
+    it('should pass when api sends successful response', (done) => {
```
ibihim: nice!

test/routes/documentRoutesTest.js
```diff
@@ -39,13 +41,10 @@ describe('documentRoutes', () => {
...
             done();
         });
-        requestStub.reset();
     });
 
     it('getFileUploadUrls passes', (done) => {
-        documentRoutes = proxyquire('../../src/routes/documentRoutes', {
-            '../lib/hcRequest': { default: requestStub.resolves('pass') },
-        }).default;
+        requestStub.resolves('pass');
```
ibihim: oh, nice.

test/routes/documentRoutesTest.js
ibihim: But you started already to name them properly in the other areas 👍

test/services/documentServiceTest.js
```diff
@@ -229,6 +234,59 @@ describe('DocumentService', () => {
...
+                .then(() => done(Error('getDocumentsCount rejection didn\'t work properly in documentService.loginNode')))
+                .catch((err) => {
+                    expect(err).to.equal('error');
+                    expect(fromFhirObjectStub).to.not.be.called;
+                    done();
+                });
+        });
+    });
+
+
```
ibihim: nice effort

src/services/FhirService.js
```diff
@@ -100,36 +89,41 @@ class FHIRService {
...
+        return Promise.all([
+            this.zeroKitAdapter.decrypt(record.encrypted_body),
+            record.encrypted_tags.map(tag => encryptionUtils.decrypt(tag, tek)),
+        ])
+            .then((results) => {
+                delete record.encrypted_body;
+                delete record.encrypted_tags;
+                record.body = JSON.parse(results[0]);
+                record.tags = results[1];
+                return record;
```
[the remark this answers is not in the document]
ibihim: As this is a nice helper function, we could consider moving it out of the class? Just a suggestion, no change request here.

But definitely the right way to move forward with this code 👍

src/routes/documentRoutes.js
```diff
@@ -22,8 +22,14 @@ const documentRoutes = {
...
 
-    searchRecords(queryParams) {
-        return hcRequest('GET', `${apiUrl}/records`, { query: queryParams, authorize: true });
+    searchRecords(userId, queryParams) {
+        return hcRequest('GET', `${apiUrl}/users/${userId}/records`, { query: queryParams, authorize: true, includeResponseHeaders: true })
+            .then(({ body, headers }) => ({ records: body, totalCount: headers['x-total-count'] }));
+    },
+
+    getRecordsCount(userId, queryParams) {
+        return hcRequest('HEAD', `${apiUrl}/users/${userId}/records`, { query: queryParams, authorize: true, includeResponseHeaders: true })
```
ibihim: why are we putting it into the header???

src/routes/documentRoutes.js
```diff
@@ -22,8 +22,14 @@ const documentRoutes = {
         return hcRequest('PUT', `${apiUrl}/users/${userId}/records/${recordId}`, { body: data, authorize: true });
     },
 
-    searchRecords(queryParams) {
-        return hcRequest('GET', `${apiUrl}/records`, { query: queryParams, authorize: true });
+    searchRecords(userId, queryParams) {
+        return hcRequest('GET', `${apiUrl}/users/${userId}/records`, { query: queryParams, authorize: true, includeResponseHeaders: true })
```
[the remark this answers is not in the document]
Harsimran1: i think it is ignoring the strings because we have this rule ```{ "code": 100, "ignoreTemplateLiterals": true, "ignoreStrings": true }```

src/routes/documentRoutes.js
Harsimran1: it is how the backend is designed. Basically when we want to search for records, it always gives the total number of counts in the headers (as we might need it if results are paginated ). If we just want to retrieve the counts of the records, we can simply call the same api with HEAD method without getting the body (which will be useless.)

test/routes/documentRoutesTest.js
Harsimran1: didn't want to mess the structure of whole file and we were anyways doing the story :)

src/services/FhirService.js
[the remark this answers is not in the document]
Harsimran1: yes, makes sense to return new object since there are only 3 values needed in the return anyway. But the function depends on zerokitadapter which is part of this class, so it will be difficult to get it out.

src/routes/documentRoutes.js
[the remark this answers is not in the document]
ibihim: oh okay... hmmm... 😄

src/routes/documentRoutes.js
ibihim: hm, okay.

src/services/FhirService.js
```diff
@@ -100,36 +89,43 @@ class FHIRService {
...
+                if (countOnly) {
+                    return documentRoutes.getRecordsCount(ownerId, queryParams);
+                }
+                return documentRoutes.searchRecords(ownerId, queryParams);
+            })
+            .then((searchResult) => {
+                totalCount = searchResult.totalCount;
+                return searchResult.records
+                    ? Promise.all(searchResult.records
+                        .map(result => this.decryptRecordAndTags(result, user.tek)))
```
ibihim: could you move the map up? if you want to have part of the line in a new line, could you do it after the fat arrow?

PR #119, opened 2018-01-23
send Version with every request to HealthCloud endpoint

Glumli:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hc-sdk-js",
-  "version": "1.0.0",
+  "version": "0.0.0",
```
Harsimran1: is this version naming in sync with Android and IOS ? I think they are using `versionName = "0.1.0"` in android.

webpack.config.js
```diff
@@ -29,9 +30,10 @@ module.exports = (env) => {
             new webpack.DefinePlugin({
                 'global.GENTLY': env.TARGET !== 'node',
                 NODE: env.TARGET === 'node',
+                VERSION: JSON.stringify(packageJson.version),
             }),
+            new webpack.BannerPlugin(`Version: ${JSON.stringify(packageJson.version)}`),
```
Harsimran1: Is there a way to add something like commit id in banner plugin? I think frank wanted it to know which SDK is deployed. But I assume you have to add commit before knowing commit id. I suppose we don't want to update version every time we commit.

package.json
Glumli: What do you mean with version naming? The current version on iOS is 0.3.0 at the moment. I just wanted to remove the 1.0.0 for now.

webpack.config.js
Glumli: We only have to update the version everytime we release. I think that's ok so far. I don't know how we want to approach Franks commit id, but I guess that's another task.

package.json
Harsimran1: i thought you changed from 0.0.0 to 1.0.0. Looks good.

ibihim: why not 0.1.0???

src/lib/hcRequest.js
```diff
@@ -34,11 +37,13 @@ const hcRequest = (type, path, {
 } = {}) => {
     let retries = 0;
 
-
     const promise = () => {
         if (authorize) {
             headers.Authorization = `Bearer ${sessionHandler.getItem('HC_Auth')}`;
         }
+        if (isHealthCloudPath(path)) {
+            headers['GC-SDK-Version'] = `JS ${VERSION}`;
```
ibihim: Didn't this trigger an eslint warning?!

webpack.config.js
ibihim: 👍

src/lib/hcRequest.js
ibihim: Maybe...?
```JavaScript
const finalHeaders = { 'GC-SDK-Version': `JS ${ VERSION } `, ...headers};
```

PR #124, opened 2018-01-24
releasing 0.1.0

ibihim:


.travis.yml
```diff
@@ -3,10 +3,6 @@ language: node_js
 cache: yarn
 services: docker
 
-branches:
-   only:
-    - master
```
ibihim: this might be the issue, that travis doesn't create releases

PR #125, opened 2018-01-26
Autoformatted code with Prettier + eslint:recommended + Airbnb + team overrides

luisfarzati:
Note: there should not be any issues, as the reformatting that Prettier does is "smart", in the sense that the source code is parsed into AST and then re-printed. BUT, I would suggest someone to do a second test locally (using the examples or Terra) before approving, just to be sure. 

Code was reformatted with Prettier using default, neutral rules (as per my initial proposal) so there's no opinionated-ism¹. On top of Prettier, eslint:recommended also provides a few autoformattable rules that are relevant.

All non-formatting rules provided either by eslint:recommended, Airbnb or us, were preserved in the eslint configuration.

¹ In my experience this is the only fair play, because even if there's a 100% agreement now, what happens when a future colleague joins and have a different opinion?


luisfarzati:
... plus for some reason the reformatted code adds 0.04% coverage 😄

luisfarzati:
Closing this as it was done as demo for Innovation Days. I'll create a new one in the future when we have time for reviewing the proposal.

PR #126, opened 2018-02-01
Return client id in documents returned

Harsimran1:


src/lib/taggingUtils.js
```diff
@@ -17,6 +17,10 @@ const taggingUtils = {
         + `${tagDelimiter}`
         + `${stringUtils.prepareForUpload(value)}`;
     },
+
+    getValue(tag) {
+        return tag.split(tagDelimiter)[1];
```
isalga: If tagDelimiter is a constant I would beter call it TAG_DELIMITER.
And what would happen if tag.split returns an array with lenght 1?

src/services/DocumentService.js
```diff
@@ -115,6 +115,8 @@ class DocumentService {
             .then((result) => {
                 result.records = result.records.map((record) => {
                     const hcDocument = hcDocumentUtils.fromFhirObject(record.body);
+                    const clientTag = record.tags.find(el => el.includes('client='));
```
isalga: If you need to know if tag includes the delimiter '=' would be better to import int from taggingUtils. Also this client tag seems a constant that can be defined and imported from somewhere else.

src/services/DocumentService.js
```diff
@@ -115,6 +115,8 @@ class DocumentService {
...
+                    hcDocument.client = clientTag ? taggingUtils.getValue(clientTag) : '';
```
isalga: What if `taggingUtils.getValue(clientTag)` return undefined?

src/services/DocumentService.js
Harsimran1: was writing a commit to import both from taggingUtils.

src/services/DocumentService.js
Harsimran1: This means we don't know the source of document. But would be ok to return the document. However I will make the return type in else case to be undefined too

src/services/DocumentService.js
isalga: 👍

src/services/DocumentService.js
isalga: 👍

src/lib/taggingUtils.js
```diff
@@ -1,6 +1,11 @@
 import stringUtils from './stringUtils';
 
-const tagDelimiter = '=';
+const TAG_DELIMITER = '=';
+
+export const tags = {
+    client: 'client',
+    documentReference: 'documentReference',
```
Glumli: What is the tag documentReference for? Isn't this a value and not a key, as 'client'?

src/services/DocumentService.js
```diff
@@ -115,6 +115,7 @@ class DocumentService {
             .then((result) => {
                 result.records = result.records.map((record) => {
```
isalga: This is not from this PR, but what if result.records is undefined or is not an array? (crash) Which is actually a possibility taking the output of `fhirService.searchRecord` ```.then(results => (results
                ? { totalCount, records: results }
: { totalCount }));```

Still encourage to use a type system.

src/lib/taggingUtils.js
Harsimran1: aah yes, sorry my bad.

src/services/DocumentService.js
Glumli: TYPE SYSTEM!!!

Harsimran1: Not a possibility because { totalCount } is returned in countOnly = true (getDocumentsCount) case, but this is the case of getDocuments. Agreed that it is visible if you check on the first look. Maybe types would be good .

isalga: Right now yes, but it would be very easy than someone refactors or makes a mistake and it would have side effects in this function ;)

PR #128, opened 2018-03-16
make sdk compatible to new auth flow

Glumli:


src/HealthCloud.js
```diff
@@ -35,50 +38,14 @@ class HealthCloud {
...
-
-        // TODO remove! This is just for not having to login after reloading
-        userService.getInternalUser()
-            .then(userService.setupUser.bind(userService))
-            .then((user) => {
-                const createUserSpecificEncryptionService =
-                        createClientEncryptionService(options.clientId)(user.CUP.privateKey);
-                documentService.setEncryptionService(createUserSpecificEncryptionService);
-            });
+    constructor({ clientId }) {
```
Glumli: We should probably kick it out

src/services/userService.js
```diff
@@ -1,52 +1,46 @@
...
 
     getUserIdForAlias(alias) {
         return this.resolveUser(alias).then(user => user.id);
     },
 
     getCurrentUser() {
-        const currentUser = sessionHandler.getItem('HC_User');
+        // TODO throw error
+        // if (!this.currentUser) throw new SetUpError(NOT_SETUP);
+        if (!this.currentUser) console.warn(new SetUpError(NOT_SETUP));
```
Glumli: This is for not breaking, but still annoying Terra

src/services/userService.js
```diff
@@ -57,18 +51,18 @@ const userService = {
...
-                    user.zeroKitId = userDetails.user.zerokit_id;
-                    user.tresorId = userDetails.user.tresor_id;
                     user.state = userDetails.user.state;
 
                     if (partial) return resolve(user);
 
                     const userDataPromise =
                         userDetails.user.user_data && userDetails.user.user_data.encrypted_data ?
-                            this.zeroKitAdapter.decrypt(userDetails.user.user_data.encrypted_data) :
+                            // TODO proper decryption
```
Glumli: removed encryption here for now. This will be put back in, in another Ticket

src/services/userService.js
```diff
@@ -93,66 +87,10 @@ const userService = {
             }));
     },
 
-    setupCUP(user) {
```
Glumli: All happening in Terra now (I guess the TEK is not specified yet, but it would make sense for me)

src/services/userService.js
```diff
@@ -171,24 +109,13 @@ const userService = {
...
                 .then(JSON.stringify)
-                .then(data => this.zeroKitAdapter.encrypt(currentUser.id, data))
+                // TODO encrypt
                 .then(encryptedUserDetails =>
                     userRoutes.updateUser(currentUser.id, { encrypted_data: encryptedUserDetails }))
                 .then(resolve)
                 .catch(reject);
         });
     },
-
```
Glumli: The grantPermissionStuff is goingt to happen in Terra as well

test/healthCloudTest.js
```diff
@@ -30,35 +30,10 @@ describe('HealthCloud', () => {
...
-        if (NODE) {
-            expect(typeof healthCloud.login).to.equal('function');
-        } else {
-            expect(typeof healthCloud.login).to.equal('undefined');
-        }
-        global.NODE = false;
-
-        done();
-    });
+    // TODO test setup
```
Glumli: Tests will be back

test/services/userServiceTest.js
```diff
@@ -24,10 +24,10 @@ describe('services/userService', () => {
     let getGrantedPermissionsStub;
     let getUserDetailsStub;
     let resolveUserIdStub;
-    let sessionHandlerGetStub;
     let updateUserStub;
 
     beforeEach(() => {
+        // TODO
         decryptStub = sinon.stub();
```
Glumli: Even though tere is no encryption happening for now it is kept in here because it's goingt to be back in near future

src/lib/hcRequest.js
```diff
@@ -1,48 +1,25 @@
...
+        query = {},
+        headers = {},
+        responseType = '',
+        authorize = false,
+        includeResponseHeaders = false,
+    } = {}) {
+        const h = headers;
         if (authorize) {
-            headers.Authorization = `Bearer ${sessionHandler.getItem('HC_Auth')}`;
+            h.Authorization = `Bearer ${this.accessToken}`;
```
Glumli: I'm not to sure about this. @luisfarzati

Glumli:
@s-urbaniak Did only delete the files for you. I would have let them in as zombieCode

src/HealthCloud.js
```diff
@@ -23,9 +21,14 @@ class HealthCloud {
         this.getUserIdByAlias = userService.getUserIdForAlias.bind(userService);
         this.getUser = userService.getUser.bind(userService);
         this.updateUser = userService.updateUser.bind(userService);
-        this.grantPermission = zeroKitAdapter.grantPermission.bind(zeroKitAdapter);
-        this.getGrantedPermissions = userService.getGrantedPermissions.bind(userService);
-        this.logout = zeroKitAdapter.logout.bind(zeroKitAdapter);
+        this.logout = userService.resetUser.bind(userService);
+
+        this.setup = (privateKey, accessToken, userId) => {
```
Glumli: I guess we should split this into one initial setupMethod and one for updating the auth token

luisfarzati: Why not these in the constructor?

src/lib/hcRequest.js
luisfarzati: LGTM, what is what you don't like?

src/HealthCloud.js
Glumli: PrivateKey and userId?
Actually a good question. I think I thought about logging out and then logging in again, then you would have to construct the sdk 2 times.

src/lib/hcRequest.js
Glumli: I don't get the Bearer prefix ;)

luisfarzati: It's going to be `Bearer` always, but can be set by the SDK consumer if you wish, sounds fine either way to me.

src/HealthCloud.js
luisfarzati: Yes, I'm all in for creating a new instance of the SDK every time this changes. We should treat this as something that is inherently attached to an instance.

Glumli: Ok and then there is another method for updating the auth token?

luisfarzati: No, I would say also in the constructor. If token expires a new instance of the SDK is created passing the new token.

src/lib/hcRequest.js
ibihim: Where would you like to put it?

src/lib/hcRequest.js
```diff
@@ -1,48 +1,25 @@
...
+
+    submit(type, path, {
+        body,
+        query = {},
+        headers = {},
+        responseType = '',
+        authorize = false,
+        includeResponseHeaders = false,
+    } = {}) {
+        const h = headers;
```
ibihim: I wouldn't call it h... this is not Go-land :smile: Especially declaring it a const and then mutating it... h is also just a reference to the reference of headers. You could use Object.assign in combination with the authorization and gc-sdk-version part.

src/lib/hcRequest.js
```diff
@@ -1,48 +1,25 @@
...
-    headers = {},
-    responseType = '',
-    authorize = false,
-    includeResponseHeaders = false,
-} = {}) => {
-    let retries = 0;
-
-    const promise = () => {
+const hcRequest = {
+    accessToken: undefined,
```
ibihim: I would rather leave it out or set it to null. Undefined is usually "untouched". "null" indicates touching from someone :smile:

src/lib/hcRequest.js
```diff
@@ -54,18 +31,9 @@ const hcRequest = (type, path, {
...
-                    retries += 1;
-                    return sendRefreshToken()
-                        .then(() => promise());
-                }
-                throw buildCustomError(err);
             });
-    };
-
-    return promise();
+        // TODO handle errors
```
ibihim: :+1:

src/services/userService.js
```diff
@@ -1,52 +1,46 @@
...
+import SetUpError, { NOT_SETUP } from '../lib/errors/SetupError';
+// import crypto from '../lib/crypto';
 import ValidationError, {
     MISSING_PARAMETERS,
     INVALID_PARAMETERS,
 } from '../lib/errors/ValidationError';
 
 const userService = {
-    user: null,
+    currentUser: undefined,
```
ibihim: Leave it out or set to null

src/services/userService.js
```diff
@@ -1,52 +1,46 @@
...
 
-    setZeroKitAdapter(adapter) {
-        this.zeroKitAdapter = adapter;
-    },
+    user: null,
 
     resetUser() {
         this.user = null;
+        this.currentUser = undefined;
```
ibihim: set to null or leave it out, not consistent with above correct handling

src/lib/errors/SetupError.js
```diff
@@ -0,0 +1,8 @@
+export default class SetupError {
```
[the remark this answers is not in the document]
ibihim: I am also a huge fan of `of` (not a change request):
```JavaScript
class CustomError extends Error {
  // evil inheritance code from hell

  static of(msg) {
    return new CustomError(msg);
  }
}

Calling new can go wrong in JS... so it is better to wrap it... in case it does not change the behavior of `of` (invalidating the stack trace).

File: src/services/userService.js (hunk 2 above)
Comment (ibihim) [reply src/services/userService.js]: woah, I am really not a fan of this... I would rather create a CustomError Factory for all the custom msgs you have :rofl:  but not a change request

... or having static functions inside of the CustomError class... :thinking:

File: src/services/userService.js (hunk 3 above)
Comment (ibihim) [reply src/services/userService.js]: Which ticket? I am afraid we lose track of things if they are just in the mind or post it of someone instead of a realy ticket

File: src/services/userService.js (hunk 2 above)
Comment (Glumli) [reply src/services/userService.js]: Of what? Of the TODO?

File: src/lib/errors/SetupError.js [hunk 16]
```diff
@@ -0,0 +1,8 @@
...
+    constructor(message) {
+        this.name = 'SetupError';
+        this.message = message || '';
+    }
+}
+
+export const NOT_SETUP = 'the sdk was not setup correctly';
```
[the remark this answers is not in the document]
Glumli: Actually I was just changin from login error, to setup error and nothing more. I thinks the first error handling was implemented by Iris. And i guess she exactly wanted to have message and class independent(e.g Loginerror(WRONG_PASSWORT) and Loginerror(WRONG_USERNAME))

src/HealthCloud.js
```diff
@@ -23,9 +21,14 @@ class HealthCloud {
...
+            documentService.setEncryptionService(
+                createClientEncryptionService(clientId)(privateKey));
```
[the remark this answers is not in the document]
Glumli: There never was a way that "createClientEncryptionService(clientId) creates privateKey variants". I wanted to refactor the whole encryptionService in https://gesundheitscloud.atlassian.net/browse/HC-1165.
I think even though refactoring is awesome, we have to find the right level of refactoring per PR.(especially when it already is quite big)

src/lib/errors/SetupError.js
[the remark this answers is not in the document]
Glumli: I planned to change the whole error stuff in https://gesundheitscloud.atlassian.net/browse/HC-809. This is code was actually not a creation of a new class, but a renaming of LoginError.
But also wanted to extend error

src/HealthCloud.js
Glumli: Already done

src/services/userService.js
Glumli: Just talked to peter. In
https://gesundheitscloud.atlassian.net/browse/HC-1096 the whole userprofile will be new. Therefore it makes no scence to work on this

src/lib/errors/SetupError.js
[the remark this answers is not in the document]
ibihim: It is not relevant who wrote this in the first place :smile:

src/lib/errors/SetupError.js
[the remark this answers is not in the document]
ibihim: A heads up on this would be cool, so I don't review code that will be changed anyway

src/services/userService.js
ibihim: This topic is part of your next story _HC-809_, I guess.

src/lib/errors/SetupError.js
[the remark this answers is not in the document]
ibihim: @Glumli please update hc-809, this story is not properly defined. It mentions the currently broken state, but does not indicate any goals or hints for what will be done, except a vague title which is already "fulfilled", by this PR.

src/lib/errors/SetupError.js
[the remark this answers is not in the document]
Glumli: I think it is important, that is wasn't me, as the person probably had a reason to do it that way. But nevermind ;)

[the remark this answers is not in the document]
luisfarzati: My 2 cents, then pls don't change it until @isalga comes back and can elaborate a bit more on the whys.

PR #129, opened 2018-03-19
implement SPKI and PKCS8

Glumli:


src/lib/crypto/index.js
```diff
@@ -18,7 +18,7 @@ const RSA_OAEP = {
 const PBKDF2 = {
     name: 'PBKDF2',
     salt: crypto.getRandomValues(new Uint8Array(16)),
-    iterations: 100,
+    iterations: 100000,
```
luisfarzati: I remember some time ago I came across this: 

> Note: LastPass supports a diverse set of platforms which vary greatly in speed. In order to utilize all of them, we recommend you do not exceed 10,000 rounds. A change from 5000 rounds to 10,000 rounds may not be perceptible to you on most platforms. However, while we permit users to increase their rounds all the way to 200,000 rounds, you may start to notice problems when logging in via certain browsers or platforms when you go above 5,000 rounds. For example, Internet Explorer 7 will be very slow with such a higher number of rounds. Logging into m.lastpass.com on a smart phone (where the rounds are done in JavaScript only) may not work at all.

I haven't digged into it but I would make sure we double check this number, especially for mobile clients. @ztepsa and/or @jzitkovic maybe already checked this out, otherwise I'd say it's worth a task in JIRA at least, to revisit this later.

Glumli: Sounds good. Will remove it from here for now.

PR #130, opened 2018-03-26
use column in sol

Glumli:


src/services/documentService.js
```diff
@@ -20,18 +20,12 @@ const documentService = {
...
-                return documentRoutes.fetchAttachmentKey(ownerId, documentId)
-                    .then((attachmentKey) => {
-                        encryptedAttachmentKey = attachmentKey;
-                        return Promise.all(
-                            hcDocument.attachments.map(attachment =>
-                                documentRoutes
-                                    .getFileDownloadUrl(ownerId, documentId, attachment.id)),
-                        );
-                    });
+                encryptedAttachmentKey = record.attachment_key;
```
ibihim: isn't the linter complaining? I would have assumed you would be forced to use `record["attachment_key"]`

Glumli: No it's not. (travis ;)) But do you like it more to call like this?

ibihim: I am just wondering. This is a necessary evil in the ajax shims, but if this is not reported, what else will be missed??

PR #132, opened 2018-03-27
H/add more fields to hcdoc

Harsimran1:
Related to https://gesundheitscloud.atlassian.net/browse/HC-1174


src/lib/models/utils/hcAttachmentUtils.js
```diff
@@ -10,14 +9,7 @@ const schema = {
     required: [],
 };
 
-const ajv = new Ajv();
-const validate = ajv.compile(schema);
-
 const hcAttachmentUtils = {
-    isValid(hcAttachment) {
```
Harsimran1: Removed this and added it as part of `documentschema`

src/services/DocumentService.js
```diff
@@ -49,11 +49,13 @@ class DocumentService {
...
         }
 
-        return this.fhirService.createFhirRecord(ownerId, hcDocumentUtils.toFhirObject(hcDocument))
-            .then((record) => {
-                hcDocument.id = record.record_id;
-                return this.updateDocument(ownerId, hcDocument);
-            });
+        return this.fhirService.createFhirRecord(ownerId, hcDocumentUtils.toFhirObject(
+            hcDocument,
+            this.zeroKitAdapter.authService.clientId),
```
Glumli: Will be a fun merge of the cryptobranch to master :(

src/lib/models/utils/hcDocumentUtils.js
```diff
@@ -1,53 +1,87 @@
...
+import hcAuthorUtils, { schema as authorSchema } from './hcAuthorUtils';
+import hcSpecialtyUtils from './hcSpecialtyUtils';
 
 const schema = {
     type: 'object',
     properties: {
         title: { type: 'string' },
-        author: { type: 'string' },
+        author: authorSchema,
+        additionalIds: {
```
Glumli: As I undnerstood additionalIds is not an array, but an Object. Where a clientId can create it's own ID
```
{
'lifetime_client_id': 'xz',
'medicalapp_client_id': 'zz',
}

Comment (Harsimran1) [reply src/lib/models/utils/hcDocumentUtils.js]: iOS seems to be using array. https://github.com/gesundheitscloud/hc-sdk-ios/blob/1e140733439c610653cb132eac679ec3762c832f/Sources/Models/FHIR/Document.swift#L45 @ztepsa any comments. And should we agree on the name? additionalIdentifiers or additionalIds ? I guess at this point it will be easier to change in js-sdk

Comment (Glumli) [reply src/lib/models/utils/hcDocumentUtils.js]: @ztepsa what do you think?
I think in https://gesundheitscloud.atlassian.net/wiki/spaces/HC/pages/3932506/hcDocument it looks like it should be a dict/object

File: src/lib/models/utils/hcDocumentUtils.js [hunk 4]
```diff
@@ -1,53 +1,87 @@
...
-            ({ attachment: hcAttachmentUtils.toFhirObject(attachment) }));
+        fhirObject.content = hcDocument.attachments ? hcDocument.attachments.map(attachment =>
+            ({ attachment: hcAttachmentUtils.toFhirObject(attachment) })) : [];
 
+        fhirObject.contained = [hcAuthorUtils.toFhirObject(hcDocument.author, clientId)];
+        if (hcDocument.additionalIds) {
+            fhirObject.identifier = hcDocument.additionalIds.map(id => ({
+                value: id,
+                assigner: {
+                    reference: id,
```
Glumli: this needs to be the clientId

src/lib/models/utils/hcDocumentUtils.js
Glumli: Talked to zoran. 
iOS is only exposing IDs those IDs that belong to the specific client. And storing the other Id's in a private var. 
As I see it here, we are exposing all IDs. 
When we are now updating a Document we would add all IDs with the updater ClientId, wouldn't we?

PR #139, opened 2018-04-13
hcReport: add model and service

luisfarzati:
Missing tests, will add those today.


src/lib/models/utils/hcReportUtils.js
```diff
@@ -0,0 +1,61 @@
+import HCReport from '../HCReport';
+import ValidationError from '../../errors/ValidationError';
+
+const toId = ({ id }) => ({ id });
+const toIdObject = id => ({ id });
+
+const hcReportUtils = {
+    isValid() {
+        return true;
```
Glumli: This validation should be implemented.

src/services/reportService.js
```diff
@@ -0,0 +1,63 @@
...
+        this.encryptionService = encryptionService;
+        this.fhirService.encryptionService = encryptionService;
+    },
+
+    downloadReport(ownerId, reportId) {
+        return this.fhirService.downloadFhirRecord(ownerId, reportId)
+            .then(record => (
+                Promise.all([
+                    record,
+                    crypto.convertBlobToArrayBufferView(record.body),
```
Glumli: convertBlobToArrayBufferView is synchron

Glumli: And is the body a blob and not a string?

src/services/reportService.js
```diff
@@ -0,0 +1,63 @@
...
+    },
+
+    uploadDocument(ownerId, hcReport) {
+        if (!hcReportUtils.isValid(hcReport)) {
+            return Promise.reject(new ValidationError('Not a valid hcReport'));
+        }
+        const fhirObject = hcReportUtils.toFhirObject(hcReport, this.clientId);
+
+        return this.fhirService.createFhirRecord(ownerId, fhirObject)
+            .then(record => this.updateDocument(ownerId, { id: record.record_id, ...hcReport }));
```
Glumli: Uh wow... uploading a record 2 times just for the id... But not your problem

src/services/reportService.js
Glumli: What  are you decrypting here in general?
This should have happened in the fhirService.

src/lib/models/utils/hcReportUtils.js
luisfarzati: Didn't we say to skip validations for now?

src/services/reportService.js
luisfarzati: `convertBlobToArrayBufferView` is async; the record body is an encrypted blob and I'm following `documentService` as example. In there, documents are decrypted by that service and not by the `fhirService`. Should I do it differently?

src/lib/models/utils/hcReportUtils.js
Glumli: The validation of the fhir yes, but I think we should still validate our own resources

luisfarzati: 👍 sure then, I'll add it.

src/services/reportService.js
Glumli: Ofcourse you're right regarding the convertion.
And of course it's also right, that the files are decrypted in the DocumentService.
But the fhirObject is handled by the fhirService. We can also sit together for clarifying

PR #141, opened 2018-04-23
updates crypto module

Glumli:


src/lib/models/utils/hcAuthorUtils.js
```diff
@@ -23,33 +23,59 @@ export const schema = {
...
+        return {
+            firstName: name.given && name.given[0] ? name.given[0] : undefined,
+            lastName: name.family,
+            prefix: name.prefix && name.prefix[0] ? name.prefix[0] : undefined,
+            suffix: name.suffix && name.suffix[0] ? name.suffix[0] : undefined,
         };
+    }
+    return {};
+};
+const hasAdress = authorFhirObject =>
```
Harsimran1: Address*

src/lib/crypto/index.js
```diff
@@ -75,13 +77,14 @@ const convertBlobToArrayBufferView = blob =>
...
+
 /**
- * Transformation from base64 encoded key to CryptoKey.
+ * Import from base64 encoded raw key to CryptoKey.
  *
  * @param {String} key - base64 encoded key
  * @returns {Promise} Resolves to the key as a CryptoKey.
  */
-const getSymKeyFromString = (key) => {
+const importSymKeyFromBase64 = (key) => {
```
Harsimran1: get seemed like a better name than import .... import feels like we're doing something more than just converting.

src/lib/crypto/index.js
```diff
@@ -135,24 +158,33 @@ const getPublicKeyFromSPKI = (SPKI) => {
     );
 };
 
-const getSPKIFromPublicKey = key =>
+
+/**
+ * Export public CryptoKey to base64 encoded SPKI.
+ *
+ * @param {Object} key - the key that should be exported
+ * @returns {Promise} Resolves to the key as an base64 encoded String.
```
Harsimran1: *`as a base64`

src/lib/crypto/index.js
```diff
@@ -135,24 +158,33 @@ const getPublicKeyFromSPKI = (SPKI) => {
...
 
+
 const importKey = (key) => {
     switch (key.t) {
     case keyTypes.COMMON_KEY:
     case keyTypes.DATA_KEY:
     case keyTypes.ATTACHMENT_KEY:
     case keyTypes.TAG_ENCRYPTION_KEY:
-        return getSymKeyFromString(key.sym);
+    case keyTypes.PASSWORD_KEY:
```
Harsimran1: does this mean that the key format is different in all these cases? base64 and object ? Would be nice to make it uniform

src/lib/crypto/index.js
```diff
@@ -164,27 +196,28 @@ const exportKey = (key, type) => {
...
     case keyTypes.TAG_ENCRYPTION_KEY:
-        return getStringFromSymKey(key)
-            .then(keyString => ({
+    case keyTypes.PASSWORD_KEY:
+        return exportSymKeyToBase64(key)
+            .then(base64 => ({
                 t: type,
                 v: 1,
-                sym: keyString,
+                sym: base64,
```
Harsimran1: does this conform with other SDKs? And is the change

src/lib/models/utils/hcAuthorUtils.js
Glumli: 👍

src/lib/crypto/index.js
Glumli: We are calling importKey. As it is more than just converting base64 to bytes import is fine I guess.

src/lib/crypto/index.js
Glumli: 👍

src/lib/crypto/index.js
Glumli: Which cases to you mean? commonKey, datakey, attachmentKey...?
https://github.com/gesundheitscloud/dev-docs/blob/master/design-documents/crypto-api.md#key-exchange-formats

src/lib/crypto/index.js
Glumli: https://github.com/gesundheitscloud/dev-docs/blob/master/design-documents/crypto-api.md#key-exchange-formats

src/lib/crypto/index.js
Harsimran1: I thought the above 5 are `base64` and the remaining asym keys are `object` types.

Glumli: the hcKey is an object. But key.sym, key.priv and key.pub are base64

src/lib/crypto/index.js
Harsimran1: I think we should agree on some naming convention for input parameters. e.g the `key` here can be called base64string since we're calling the same at other places. Since we don't have typescript right now, its difficult to guess what kind of structure is expected without seeing comments in the module,

src/lib/crypto/index.js
```diff
@@ -118,12 +127,26 @@ const getPrivateKeyFromPKCS8 = (PKCS8) => {
...
     crypto.subtle.exportKey('pkcs8', key)
         .then(PKCS8 => new Uint8Array(PKCS8))
         .then(convertArrayBufferViewToBase64);
 
-const getPublicKeyFromSPKI = (SPKI) => {
+
+/**
+ * Import public CryptoKey from base64 encoded SPKI.
+ *
+ * @param {Object} key - the base64 encoded SPKI
```
Harsimran1: Similarly, this seems like base64 string, but param says object. You can also rename the input `param to base64SPKI`

src/lib/crypto/index.js
Glumli: I can rename it. but this ticket is not actually about changing anything, but simply about adapting the changes made in Terra and actually we got the params definded above.

src/lib/crypto/index.js
```diff
@@ -95,18 +98,24 @@ const getSymKeyFromString = (key) => {
...
     crypto.subtle.exportKey('raw', key)
         .then(byteKey => new Uint8Array(byteKey))
         .then(convertArrayBufferViewToBase64);
 
 
-const getPrivateKeyFromPKCS8 = (PKCS8) => {
+/**
+ * Import private CryptoKey from base64 encoded PKCS8.
+ *
+ * @param {Object} key - the base64 encoded PKCS8 string
```
Harsimran1: Can you change the `Object` to `string` since it is added in this PR

PR #142, opened 2018-04-24
Js/enabling auth

Glumli:
implements [HC-1351](https://gesundheitscloud.atlassian.net/browse/HC-1351) and prepares the sdk to be easier usable for 3rd Party Apps.
Of course I would also update in Terra.
How the SDK is going to be used can be seen here in [PR-140](https://github.com/gesundheitscloud/hc-sdk-js/pull/140).


example/tabs/login.js
```diff
@@ -2,10 +2,9 @@
...
-        loginResultElement.innerHTML = `Successfully logged in: ${JSON.stringify(success)}`;
-    })
-    .catch((error) => {
-        loginResultElement.innerHTML = `An error occurred: ${JSON.stringify(error)}`;
-    });
+async function login() {
+    console.log('login started');
+    let clientId = 'example';
+    let CAP = await HC.healthCloud.createCAP();
+    HC.authCloud.login(clientId, CAP);
```
Glumli: AuthCloud is going to be in an extra repository

src/healthCloud.js
```diff
@@ -0,0 +1,50 @@
...
+    getCurrentUser: userService.getCurrentUser.bind(userService),
+    getUserIdByAlias: userService.getUserIdForAlias.bind(userService),
+    getUser: userService.getUser.bind(userService),
+    updateUser: userService.updateUser.bind(userService),
+    logout: userService.resetUser.bind(userService),
+    updateAccessToken: hcRequest.setAccessToken.bind(hcRequest),
+    createCAP: () =>
+        hcCrypto.generateAsymKeyPair(hcCrypto.keyTypes.APP).then(({ publicKey, privateKey }) => ({
+            publicKey: atob(JSON.stringify(publicKey)),
+            privateKey: btoa(JSON.stringify(privateKey)),
```
Glumli: exports the keys base64 encoded so they are easier to query.

src/healthCloud.js
```diff
@@ -0,0 +1,50 @@
...
+
+    models: {
+        HCDocument,
+        HCAttachment,
+        HCAuthor,
+        HCSpecialty,
+    },
+
+    setup(clientId, userId, base64PrivateKey, requestAccessToken) {
+        const privateKey = JSON.parse(atob(base64PrivateKey));
```
Glumli: Expects publicKey base 64 encoded, so nothing needs to be changed after creating them.

src/lib/models/HCAttachment.js
```diff
@@ -22,25 +22,6 @@ class HCAttachment {
         this.creationDate = creationDate || (file ? file.lastModifiedDate : undefined);
         this.id = id;
     }
-
-    static fromFhirObject(fhirObject) {
```
Glumli: This was still here. Don't know why.

.eslintrc.js
```diff
@@ -18,10 +18,10 @@ module.exports = {
         'newline-per-chained-call': ['error', { ignoreChainWithDepth: 3 }],
         'max-len': ['error', { code: 100, ignoreTemplateLiterals: true, ignoreStrings: true }],
 
-        semi: 'error',
+        'semi': 'error',
         'key-spacing': ['error', { beforeColon: false }],
 
-        complexity: ['error', 15], // push down to 10
+        'complexity': ['error', 15], // push down to 10
```
ibihim: why?

Glumli: consistency ;)

example/index.html
```diff
@@ -37,9 +37,7 @@
     </div>
 
     <script type="application/javascript">
-        var HC = new healthcloud_sdk({
-            "clientId":clientId
-        });
+        var HC = healthcloud_sdk;
```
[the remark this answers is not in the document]
Glumli: What would be a good way? simply using the object?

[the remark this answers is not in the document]
ibihim: in our library, we should have something like that:
```JavaScript
(function (factory) {
  "use strict";
  var healthcloud = factory();

  if (typeof window !== "undefined") {
    window.Healthcloud = healthcloud;
  } else {
    this.Healthcloud = healthcloud;
  }

  // requireJS
  if (typeof define === "function" && define.amd) {
    define("Healthcloud", [], healthcloud);
  }
}(function (undefined) {
  "use strict";
  
  // our bundled stuff
  
 return Healthcloud // function that is factory mentioned above
}))
```

Then the 3rd part implements it like:
```JavaScript
(function () {
  // 3rd parties namespace world
  var hc = Healthcloud.create(config); // or Healthcloud.create(param1, param2,...);
  // ...
)();
```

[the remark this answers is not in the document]
Glumli: What about using a Class? :D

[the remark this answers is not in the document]
ibihim: To have even a better implementation, that would indicate that we have more than one JS lib, we would need to do also something like that:
```JavaScript
window.Healthcloud = Healthcloud || {};
window.Healthcloud.SDK = factory();
```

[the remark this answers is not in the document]
ibihim: We should have at least a story for that.

example/tabs/login.js
ibihim: look above:
```JavaScript
window.Healthcloud = Healthcloud || {};
window.Healthcloud.SDK = sdk(config);
window.Healthcloud.AUTH = auth(config);
```

example/index.html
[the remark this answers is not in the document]
ibihim: Would it be possible, within major extend of your working time to... make the lib work something like this...
```JavaScript
var hc = Healthcloud.SDK.setup(/* params */);
```
?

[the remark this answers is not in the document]
ibihim: And please create a story for integration with different depednency systems (requirejs / nodejs / ... )

example/tabs/login.js
Glumli: It's actually still not save how we are going to integrate the auth stuff. But it might be the way to go

example/index.html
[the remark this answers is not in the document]
Glumli: https://gesundheitscloud.atlassian.net/browse/HC-1422

.eslintrc.js
ibihim: lol. :grinning:

example/tabs/login.js
```diff
@@ -2,10 +2,9 @@
...
-    .then((success) => {
-        loginResultElement.innerHTML = `Successfully logged in: ${JSON.stringify(success)}`;
-    })
-    .catch((error) => {
-        loginResultElement.innerHTML = `An error occurred: ${JSON.stringify(error)}`;
-    });
+async function login() {
+    console.log('login started');
+    let clientId = 'example';
+    let CAP = await GC.SDK.createCAP();
```
ibihim: why are clientId and CAP lets?

src/healthCloud.js
```diff
@@ -0,0 +1,50 @@
...
+import hcRequest from './lib/hcRequest';
+import taggingUtils from './lib/taggingUtils';
+import userService from './services/userService';
+import HCDocument from './lib/models/HCDocument';
+import HCAttachment from './lib/models/HCAttachment';
+import HCAuthor from './lib/models/HCAuthor';
+import HCSpecialty from './lib/models/HCSpecialty';
+
+
+const healthCloud = {
```
ibihim: is healthCloud still the proper name?

.eslintrc.js
ibihim: there is a seed of truth in it

Glumli: even though removing the 's everywhere might have been the way to go ;)

example/tabs/login.js
Glumli: awesome question, and why does the linter allow this. Ah but it's the example app. Everything is allowed in the example app

src/healthCloud.js
Glumli: Uff... this is a big one... in theory it would be SDK then, but I still like healthCloud as a name for now. It's so awsome historical.

PR #143, opened 2018-04-24
switch from aes-cbc to aes gcm

Glumli:


src/lib/crypto/index.js
```diff
@@ -224,23 +224,33 @@ const exportKey = (key, type) => {
...
  * Symmetric encryption of data with hcKey
  *
  * @param {Object} hcKey - The hcKey representation of the key
  * @param {ArrayBufferView} data that should be encrypted
  * @returns {Promise} Resolves to encrypted data as an ArrayBufferView
  */
-const symEncrypt = (hcKey, data) =>
-    importKey(hcKey)
+const symEncrypt = (hcKey, data) => {
+    const iv = hcKey.t === keyTypes.TAG_ENCRYPTION_KEY ?
```
Glumli: Had to do this, because we need deterministic encryption for tags.

ibihim: hmmmmmmmmmmmmmmmmmm...........

src/lib/crypto/index.js
```diff
@@ -261,14 +271,17 @@ const symEncryptBlob = (hcKey, blob) =>
  * @param {ArrayBufferView} encrypted data that should be decrypted
  * @returns {Promise} Resolves to plain data as an ArrayBufferView
  */
-const symDecrypt = (hcKey, data) =>
-    importKey(hcKey)
+const symDecrypt = (hcKey, ivData) => {
+    const iv = ivData.slice(0, 16);
+    const data = ivData.slice(16, ivData.length);
```
ibihim: :+1: for using slice.

src/lib/crypto/index.js
```diff
@@ -312,7 +325,7 @@ const asymEncrypt = (hcPublicKey, data) =>
         .then(key => crypto.subtle.encrypt(
             {
                 name: key.algorithm.name,
-                iv,
+                IV,
```
ibihim: and why are we using the IV here?

Glumli: It is used for the padding.

PR #144, opened 2018-04-24
Get common_key and tek and decrypt

Harsimran1:


src/services/userService.js
```diff
@@ -24,28 +26,63 @@ const userService = {
...
-        // if (!this.currentUser) throw new SetUpError(NOT_SETUP);
-        if (!this.currentUser) console.warn(new SetUpError(NOT_SETUP));
-
+    setPrivateKey(privateKey) {
+        this.privateKey = privateKey;
+    },
 
-        return this.currentUser;
+    getCurrentUser() {
+        return new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
Harsimran1: this is to specify the return type just in beginning of the function. So that no mistakes are commited while changing or implementing the function. Less cleaner, but good in long run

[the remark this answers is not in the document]
Glumli: Sounds fair

src/services/userService.js
```diff
@@ -24,29 +25,59 @@ const userService = {
...
+            if (this.user) {
+                resolve(this.user);
+                return;
+            }
+            this.setCurrentUser(this.privateKey)
```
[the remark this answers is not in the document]
ibihim: This naming is wrong.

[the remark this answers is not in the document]
ibihim: This is bad. You are setting something in a get function and also depending on the result of the set function in it.

I hope the renaming makes it less an issue / clarifies the intent...

src/services/userService.js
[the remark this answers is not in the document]
ibihim:
```JavaScript
    getCurrentUser() {
        return new Promise((resolve, reject) => {
            if (this.user) {
                resolve(this.user);
                return;
            }
            this.setCurrentUser(this.privateKey)
                .then(resolve)
                .catch(reject);
        });
},
```
vs
```JavaScript
    getCurrentUser() {
        return this.user
            ? Promise.resolve(this.user)
            : this.setCurrentUser(this.privateKey);
    },
```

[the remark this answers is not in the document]
Harsimran1: @ibihim I am fan of making code compact. But in the second scenario:
I have to go inside this.setCurrentUser to check what it returns . Since no typescript here, a given function can just return anything. Only upto developer discipline to make sure both methods return the same thing. But I am open to making changes.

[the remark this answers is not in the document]
ibihim: @Harsimran1 JS is all about developer discipline. I know TypeScript is a safety net for big companies how have a high chance to get developers that are below average... but I am pretty confident, that this is not the case at this company. If you like you could add a JSDoc comment to the top, so some IDE diggest that information and return warnings on wrong types.
And it is NOT necessary to look into this.setCurrentUser (which is now obviously bad naming) as you see a Promise.resolve one line above. This would be really strange, if someone would just return a string instead... and no one would register that in a PR review.

For me it is not compact code vs explicit code. It is redundant code due to lack of trust in "developer discipline".

src/services/userService.js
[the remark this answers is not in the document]
ibihim: The idea, to make it more explicit what happens *within the function* with `() => resolve(this.user)` is the right way, but I am curious, where in the initialization we fail to set the privateKey *outside of the function*.

src/services/userService.js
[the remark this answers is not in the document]
ibihim: If you consider changes, it would be nice to follow what I said in the next comment. In case you need help, please reach out to me :slightly_smiling_face:

src/services/userService.js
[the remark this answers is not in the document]
ibihim: I think the problem within the service is that it is not a real service. It is a huge thingy. It is a model and a service intertwined.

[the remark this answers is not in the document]
ibihim: If it was me (what I assume), I think I made a lousy job in transforming it from a class to a POJSO :smiley:

src/services/userService.js
```diff
@@ -24,29 +25,58 @@ const userService = {
...
+                .catch(reject);
+        });
     },
 
     isCurrentUser(userId) {
-        return userId === this.getCurrentUser();
+        return userId === this.currentUserId;
+    },
+
+    setCurrentUser(privateKey) {
```
ibihim: maybe fetchUser?

ibihim: and privateKey is somehow already set?

ibihim: I mean, "Das Kind ist schon in den Brunnen gefallen", it would be beyond scope to refactor stuff. 

I am not happy with the solution, but something like
```
        return new Promise((resolve, reject) => {
            if (this.user) {
                resolve(this.user);
                return;
            }
            this.fetchUsert()
                .then(() => resolve(this.user))
                .catch(reject);
        });
```
would be okay.

ibihim: translation of the meaning of "das Kind ist bereits in den Brunnen gefallen":
to close the barn door after the horse has escaped

Harsimran1: privateKey is set on initialisation of SDK.

ibihim: if it is set on initalization and there is a this.privateKey, why do we need to give it as particular arg for the function? just curious

Harsimran1: no we don't, changed it in the new adaptation. Initially I was thinking to use setCurrentUser(privateKey) on initialisation , then realised it would become a promise, so decided against it.

src/routes/userRoutes.js
```diff
@@ -14,6 +14,10 @@ const userRoutes = {
         return hcRequest.submit('GET', `${apiUrl}/users/${userId}`, { authorize: true });
     },
 
+    userInfo() {
```
Glumli: fetchUserInfo would be nicer JS naming. But can stay like this for now.

src/services/userService.js
```diff
@@ -24,29 +26,52 @@ const userService = {
...
     },
 
     isCurrentUser(userId) {
-        return userId === this.getCurrentUser();
+        return userId === this.currentUserId;
+    },
+
+    fetchCurrentUser() {
+        if (!this.privateKey) {
+            throw new SetUpError(NOT_SETUP);
```
Glumli: 👍

src/routes/userRoutes.js
ibihim: ufff

src/services/userService.js
```diff
@@ -24,29 +26,52 @@ const userService = {
...
+            .then(res => crypto.asymDecryptString(this.privateKey, res.common_key)
+                .then(JSON.parse)
+                .then((key) => {
+                    commonKey = key;
+                    return crypto.symDecryptObject(commonKey, res.tag_encryption_key);
+                })
+                .then((tek) => {
+                    this.currentUserId = res.sub;
+
+                    this.user = {
```
ibihim: well this function does do now two things:
 - fetching a user
 - setting the user as current state

you know what, what about naming it pullCurrentUser? like git logic. if you do a fetch, you wont break your banch... but a pull might create merge conflicts...

src/routes/userRoutes.js
ibihim: please change it.

src/services/userService.js
```diff
@@ -24,29 +26,52 @@ const userService = {
...
+            );
     },
 
     getInternalUser(userId, partial = false) {
-        return new Promise((resolve, reject) => {
-            const id = userId || this.getCurrentUser();
-            if (this.isCurrentUser(id) && this.user) {
-                return resolve(this.user);
-            }
+        if (!(userId && this.currentUserId) || (userId === this.currentUserId)) {
```
ibihim: hey, maybe I am not too much into this codebase, but do you think, you could pack this 2 boolean results into variables to have it easy to understand? i know this might sound funny from someone who appreciates fp :stuck_out_tongue_winking_eye: 
```if (!hasUserIds || isOwnId)```
Did I get it right?

src/services/userService.js
ibihim: Sorry for the permanent name change suggestion, the initial problem is a bigger design fault. Just trying to fix design by naming ;)

PR #145, opened 2018-04-27
Add annotations handling to documentService.updateDocument

Harsimran1:


src/lib/taggingUtils.js
```diff
@@ -17,6 +19,10 @@ const taggingUtils = {
             this.buildTag(tagKey, tagObject[tagKey]));
     },
 
+    generateCustomTags(annotationList) {
+        return annotationList.map(el => `${ANNOTATION_LABEL}${el}`);
```
Glumli: Why not `${ANNOTATION_LABEL}${TAG_DELIMITER}${el}`?

src/services/documentService.js
```diff
@@ -132,6 +138,7 @@ const documentService = {
                     const hcDocument = hcDocumentUtils.fromFhirObject(record.body);
                     hcDocument.client = taggingUtils.getTagValueFromList(record.tags,
                         tagKeys.client);
+                    hcDocument.annotations = taggingUtils.getAnnotaions(record.tags);
```
Harsimran1: *annotations

src/services/fhirService.js
```diff
@@ -6,34 +6,46 @@ import dateUtils from '../lib/dateUtils';
...
 const fhirService = {
-    updateFhirRecord(ownerId, recordId, fhirObject, attachmentKey = null) {
+    updateFhirRecord(ownerId, recordId, fhirObject, customTags = [], attachmentKey = null) {
         const updateRequest =
             (userId, params) => documentRoutes.updateRecord(userId, recordId, params);
 
         return this.downloadFhirRecord(ownerId, recordId)
             .then((record) => {
+                const tags = [
+                    ...new Set([
```
Glumli: Why do you use a set here? to not have dublications? I would actually just remove the old ones. As we have annotations as a part of the document now, we should tread them like every other attribute of it.

src/lib/taggingUtils.js
```diff
@@ -31,6 +37,14 @@ const taggingUtils = {
     getValue(tag) {
         return tag.split(TAG_DELIMITER)[1];
     },
+
+    filterCustomTags(tagList) {
```
Glumli: filterCusomTags for me sounds as if you want to remove the customTags.

src/services/fhirService.js
```diff
@@ -6,34 +6,46 @@ import dateUtils from '../lib/dateUtils';
...
     createFhirRecord(ownerId, fhirObject) {
         return this.uploadFhirRecord(ownerId, fhirObject, documentRoutes.createRecord);
     },
 
-    uploadFhirRecord(ownerId, fhirResource, uploadRequest, attachmentKey) {
+    uploadFhirRecord(ownerId, fhirResource, uploadRequest, customTags = [], attachmentKey) {
         return fhirValidator.validate(fhirResource)
             .then(() => {
-                const tags = [
+                const tagList = [
```
Glumli: why the renaming?

test/services/documentServiceTest.js
```diff
@@ -294,6 +295,36 @@ describe('documentService', () => {
                 .catch(done);
         });
 
+        it('should resolve, and pass annotations to updateFhirRecord', (done) => {
```
Glumli: Is there any value in having this test and the previous one seperated?

src/services/fhirService.js
Harsimran1: But we are merging the fhirObject. But since attachments being an array in hcDocument are also updated this way, makes sense to just update with the latest ones

test/services/documentServiceTest.js
```diff
@@ -68,7 +68,7 @@ describe('documentService', () => {
         title,
         type,
     };
-    const attachmentWithFile = JSON.parse(JSON.stringify(attachmentWithoutFile));
```
Glumli: :(

src/lib/taggingUtils.js
```diff
@@ -17,6 +19,10 @@ const taggingUtils = {
             this.buildTag(tagKey, tagObject[tagKey]));
     },
 
+    generateCustomTags(annotationList) {
+        return annotationList.map(el => this.buildTag(ANNOTATION_LABEL, el));
```
ibihim: :+1: for using map :smile: 
just imagine how it could look like, if this.buildTag would be curried? 
```JavaScript
return annotationList.map(this.buildTag(ANNOTATION_LABEL);
```
oh my... my functional phantasies start to spin, let me stop before introducing compose! :smile: 

not a change requirement

src/lib/taggingUtils.js
```diff
@@ -31,6 +37,11 @@ const taggingUtils = {
     getValue(tag) {
         return tag.split(TAG_DELIMITER)[1];
     },
+
+    getAnnotations(tagList) {
+        return tagList.filter(el => el.includes(`${ANNOTATION_LABEL}${TAG_DELIMITER}`))
+            .map(this.getValue);
```
ibihim: even though this is pretty pretty, iterating twice over the array could be expansive. as i don't know how big the filtered result, I would avoid that. Prove me wrong, if this is a wrong statement.

```JavaScript
return tagList.reduce((sum, value) => {
    if (el.includes(`${abc}${def}`)) {
        sum.push(value);
        return sum;
    }
    
    return sum;
}, [])

File: src/services/fhirService.js [hunk 10]
```diff
@@ -6,34 +6,41 @@ import dateUtils from '../lib/dateUtils';
...
+                    ownerId, updatedFhirObject, updateRequest, customTags, attachmentKey);
             });
     },
 
     createFhirRecord(ownerId, fhirObject) {
         return this.uploadFhirRecord(ownerId, fhirObject, documentRoutes.createRecord);
     },
 
-    uploadFhirRecord(ownerId, fhirResource, uploadRequest, attachmentKey) {
+    uploadFhirRecord(ownerId, fhirResource, uploadRequest, customTags = [], attachmentKey) {
```
ibihim: So one needs to call uploadFhirRecord like this:
```JavaScript
uploadFhirRecord(ownerId, fhirResource, uploadRequest, null, attachmentKey)
```
Would it make sense to swap places of attachmentKey and customTags?

test/services/documentServiceTest.js
ibihim: haha, nice cloning :smile:

test/services/documentServiceTest.js
```diff
@@ -68,7 +68,7 @@ describe('documentService', () => {
...
+    const attachmentWithFile = Object.assign({}, attachmentWithoutFile);
```
ibihim: How deep does `attachmentWithoutFile` goes? Object.assign is not a deep copy. It is legit to use it under the assumption that the **deeper levels** of the object never change... but this not functional code, I am curious... :smile:

test/services/documentServiceTest.js
```diff
@@ -80,7 +80,7 @@ describe('documentService', () => {
         attachments: [attachmentWithoutFile],
     };
 
-    const hcDocumentWithFileData = JSON.parse(JSON.stringify(hcDocumentWithoutFileData));
+    const hcDocumentWithFileData = Object.assign({}, hcDocumentWithoutFileData);
```
ibihim: Same as above

test/services/documentServiceTest.js
ibihim: I am raising this question, as someone wrote the initial code with deep copying in mind.

src/lib/taggingUtils.js
ibihim: So if you tell me that the result of filter is very small, it doesn't matter.

test/services/documentServiceTest.js
Glumli: Uh nice... now I know, why it did the upper thing :D

src/lib/taggingUtils.js
ibihim: You could consider using a for loop, but I avoid them as hell, due to my nodejs background and async :smile:

ibihim:
```JavaScript
(tagList) => {
  const result = []
  for (let i = tagList.length; i--;) {
    let el = tagList[i]
    if (el.includes()) {
      result.push(tagList[i]) // you could end up with an array full of tagList[0]s or none at all in nodejs :smile: 
    }
  }
}
```

ibihim: nothing I would like to do :smile:

test/services/documentServiceTest.js
ibihim: parse / stringify is kinda costly, so Object.assign would make a lot of sense, if we have immutability.

Harsimran1: `attachmentWithoutFile ` is something like this. `{title: "abc", type: "abc", creationDate: Date()}`   parse/stringify may have issues with date type though.

test/services/documentServiceTest.js
Harsimran1: In case of `hcDocumentWithoutFileData` we have 2 levels `{author: {firstName: "lastName"}, creationDate: Date();
}` in parse/stringify we are explicilty setting back date to original type. Currently author is never really changed(we may have test  later on). We might want to keep it in parse/stringify fashion in case we want to add tests later on...

src/lib/taggingUtils.js
```diff
@@ -31,6 +37,15 @@ const taggingUtils = {
     getValue(tag) {
         return tag.split(TAG_DELIMITER)[1];
     },
+
+    getAnnotations(tagList) {
+        return tagList.reduce((annotations, el) => {
+            if (el.includes(`${ANNOTATION_LABEL}${TAG_DELIMITER}`)) {
+                annotations.push(this.getValue(el));
+            }
+            return annotations;
```
ibihim: you improved my orginal code very well :+1:

test/services/documentServiceTest.js
ibihim: documentServiceTest.js is a test file is it? ... ship it like it is!

src/services/fhirService.js
```diff
@@ -6,34 +6,41 @@ import dateUtils from '../lib/dateUtils';
...
+                    ownerId, updatedFhirObject, updateRequest, customTags, attachmentKey);
             });
     },
 
     createFhirRecord(ownerId, fhirObject) {
         return this.uploadFhirRecord(ownerId, fhirObject, documentRoutes.createRecord);
     },
 
-    uploadFhirRecord(ownerId, fhirResource, uploadRequest, attachmentKey) {
+    uploadFhirRecord(ownerId, fhirResource, uploadRequest, customTags = [], attachmentKey = null) {
```
ibihim: I see now

PR #148, opened 2018-05-02
src/**/user*: implement getReceivedPermissions

Glumli:


test/routes/userRoutesTest.js
```diff
@@ -1,12 +1,15 @@
 /* eslint-env mocha */
 /* eslint-disable no-unused-expressions */
+/* eslint-disable max-nested-callbacks */
```
Harsimran1: is it really used in this file ? can't find ..

Glumli: In which file? When not ignoring it here it complains because of
describe
  describe
    it
      then

Harsimran1: oh ok, i meant userRoutesTest.js.

Glumli: What is not used in this file?

Harsimran1: The `eslint-disable max-nested-callback` rule . It is used, couldn't figure out where are so many callbacks. Seems like a case in most of test files. We can consider changing the configuration to 4 rather than 3

Glumli: I think we should have an extra eslint file for the test folder. I don't care how many nestings we have in test, but in the src folder 3 is a good thing

PR #149, opened 2018-05-02
src/services/userService: implements grantPermission

Glumli:


src/services/userService.js
```diff
@@ -97,6 +98,45 @@ const userService = {
...
                 })));
     },
+
+    /**
+     * @param {String} appId - the id of the user-client tuple the permission shall be granted to
+     * @param {Array} annotations - the annotations that shall be shared.
+     * @returns {Promise}
+     */
+    grantPermission(appId, annotations = []) {
+        const scope = ['rec:r', 'rec:w', 'attachment:r', 'attachment:w', 'user:r', 'user:w', 'user:q'];
```
Harsimran1: Shouldn't `'rec:r', 'rec:w', 'attachment:r', 'attachment:w'` suffice in this case ?

test/lib/cryptoTest.js
```diff
@@ -93,8 +81,8 @@ describe('crypto', () => {
         describe('asymDecryptString', () => {
             it(`should decrypt ${cipherText} and ${cipherResult} to ${plainText}`, (done) => {
                 Promise.all([
-                    hcCrypto.asymDecryptString(privHCKey, cipherText),
-                    hcCrypto.asymDecryptString(privHCKey, cipherResult),
+                    hcCrypto.asymDecryptString(encryptionResources.hcPrivateKey, cipherText),
+                    hcCrypto.asymDecryptString(encryptionResources.hcPrivateKey, cipherResult),
```
Harsimran1: It seems like this test is dependent on the previous one ....

src/services/userService.js
Glumli: Was not sure... wanted to rather grant to much for now. I also wrote everything the way, that we can simply request the scopes as a parameter.
@pierreprinetti what do you think are the permissions the other party might need?

test/lib/cryptoTest.js
Glumli: Yes the crypto tests are, but that's not a part of the PR. But there is a ticket for writing proper tests for the crypto module.

src/services/userService.js
pierreprinetti: It looks like they should suffice indeed.

https://github.com/gesundheitscloud/dev-docs/blob/master/design-documents/login-management.md#scopes

PR #150, opened 2018-05-04
Js/multiple fixes

Glumli:


src/lib/hcRequest.js
```diff
@@ -42,23 +42,25 @@ const hcRequest = {
...
     } = {}) {
         let retries = 0;
-        const headers = givenHeaders;
+        const httpHeaders = headers;
 
         const accessTokenPromise = authorize ? this.getAccessToken(ownerId) : Promise.resolve(null);
 
         if (isHealthCloudPath(path)) {
-            headers['GC-SDK-Version'] = `JS ${VERSION}`;
+            httpHeaders['GC-SDK-Version'] = `JS ${VERSION}`;
```
Harsimran1: does `GC-SDK-Version` header works with userinfo endpoint? It says header not allowed for me

src/lib/hcRequest.js
```diff
@@ -42,23 +42,25 @@ const hcRequest = {
...
         }
         const submitRequest = accessToken => request(type, path)
-            .set(headers)
-            .set('Authorization', accessToken)
+            .set({
+                ...httpHeaders,
+                authorization: accessToken || undefined,
```
Harsimran1: We might need capitalised `Authorization` ?

Glumli:
@ibihim Feel free to rewrite.

src/services/documentService.js
```diff
@@ -80,16 +80,14 @@ const documentService = {
             }
         });
 
-        // will resolve to undefined if no attachmentKey is set
-        const encryptedAttachmentKeyPromise =
-            documentRoutes.fetchAttachmentKey(ownerId, hcDocument.id);
+        const recordPromise =
+            documentRoutes.downloadRecord(ownerId, hcDocument.id);
```
Harsimran1: hmmm seems like we're calling the same api  twice in updateDocument...

src/lib/hcRequest.js
Glumli: Ah I'm working with localhost, that's why it is working for me. But I guess this should to be added to vega then, shouldn't it? I would suggest to remove it for now then?

src/services/documentService.js
Glumli: Maybe we should add a toDo, but this should not stay like this in the future. We had to do it on now, because an attachmenntKey endpoint has no prio, but in general we should definitly have one

src/lib/hcRequest.js
Glumli: I think we don't didn't have any issues

Harsimran1: And won't it be `undefined` by default ?

Glumli: I think accessToken would be null. But er can check if it also works without

src/services/documentService.js
Harsimran1: If we're anyways using downloadRecord api later, we might want to utilise it for getting the attachmentKey too rather than calling `attachment` api?  We can think about a refactor later. I will add a ticket for this.

Harsimran1: https://gesundheitscloud.atlassian.net/browse/HC-1463

PR #155, opened 2018-05-07
Ko/add auth for icarus pretty

ibihim:


src/authCloud/index.js
```diff
@@ -0,0 +1,76 @@
...
+
+const config = ({ clientId, clientURL, gcSDK }) => {
+    const postKeys = ({ privateKey, publicKey }) =>
+        request.post(`${clientURL}/gckeys`)
+            .send({ private_key: privateKey, public_key: publicKey })
+            .then(identity);
+
+    const getToken = () =>
+        request.get(`${clientURL}/gctoken`).then(({ text }) => {
+            const resObj = JSON.parse(text) || {};
```
Glumli: Did you check if there is a body?

example/server.js
```diff
@@ -1,10 +1,20 @@
...
-let port = 8888;
+const express = require('express');
+const cors = require('cors');
+const proxy = require('http-proxy-middleware');
+const app = express();
+const host = '0.0.0.0';
+const port = 8888;
 
 app.use('/', express.static('./'));
+app.use('/icarus', proxy({
```
theintz: I would suggest not to use the cryptic (for outsiders) names as routes.

src/authCloud/index.js
```diff
@@ -0,0 +1,76 @@
...
+                onceElse(
+                    () => Promise.resolve(token),
+                    getToken,
+                ),
+            );
+
+            return true;
+        }
+
+        cap.then(postKeys).then(console.log);
```
Glumli: you wanted to rename this

Glumli: console.log

example/server.js
```diff
@@ -1,10 +1,30 @@
...
+const https = require('https');
+const cors = require('cors');
+const proxy = require('http-proxy-middleware');
+const host = '0.0.0.0';
+const port = 8888;
 
-let app = express();
-let host = '0.0.0.0';
-let port = 8888;
+const privateKey = fs.readFileSync('sslcert/server.key', 'utf8');
```
Glumli: Please don't make the keys required. Let's check if we really need those for local development

src/authCloud/index.js
Glumli: ok fail

src/authCloud/index.js
```diff
@@ -0,0 +1,76 @@
...
+            sdk.setup(
+                clientId,
+                privateKey,
+                onceElse(
+                    () => Promise.resolve(token),
+                    getToken,
+                ),
+            );
+
+            return true;
```
Glumli: return true after setup, but not in parallel.

PR #157, opened 2018-05-14
implements getCurrentAppId

Glumli:
Not manually tested yet. Would suggest not to change the tests, as it will be removed in half a month.


Glumli:
@ibihim PTAL

PR #164, opened 2018-05-18
adds fhir-schema to sdk

Glumli:


Harsimran1:
Why ? This will drastically increase the size of SDK.  Is it temporary?  I suggest we host the schema somewhere and not add in SDK.

Glumli:
@Harsimran1  https://gesundheitscloud.atlassian.net/browse/HC-1510
Tobi asked for it.. I guess at least for the demo this is ok.

theintz:
Does this add the *entire* FHIR schema? Or just the bits that we need?

Harsimran1:
It adds the entire FHIR schema. but when loaded over API , it reduces down to 30KB (on first request, and later loaded from disc cache).

PR #172, opened 2018-06-15
Fix document sorting

Harsimran1:
https://gesundheitscloud.atlassian.net/browse/HC-1600


src/lib/models/HCDocument.js
```diff
@@ -7,6 +7,8 @@ class HCDocument {
         type = 'Document',
         title,
         creationDate = new Date(),
+        updatedDate,
```
Glumli: Should we really expose all of them? I have the feeling that only one date should be exposed. The date on which the user says the Document was created. I always thought the others should only be set by us. Maybe exposing an update date might make sense as well, but I don't really think so.

src/services/documentService.js
```diff
@@ -138,10 +138,12 @@ const documentService = {
                 hcDocument.attachments = [...oldAttachments, ...newAttachments];
                 return this.fhirService.updateFhirRecord(
                     ownerId,
-                    hcDocument.id,
-                    hcDocumentUtils.toFhirObject(hcDocument, this.clientId),
-                    customTags,
-                    attachmentKey,
+                    {
```
Glumli: I don't think that this is better than before. For me it looks like it could make sense to bundle all record related stuff in one(but why not putting the ownerId in the object than as well). On the other side it makes things look worse often as well. Also I would prefer a more meaningful name than 'body' for the fhirResource(might be a great name btw ;) ) itself.
 
Will continue thinking about it, but also would like to have @ibihim opinion.

src/lib/models/HCDocument.js
Harsimran1: Yes these fields should not be exposed to user to update. Will remove creation and updatedDate.

src/services/documentService.js
Glumli: Ok... then go for it ;)

src/services/fhirService.js
```diff
@@ -8,46 +8,83 @@ import createCryptoService from './cryptoService';
...
+     * @param {Object} record
+     * @param {string} record.id
+     * @param {Object} record.body The Fhir Resource contained in record
+     * @param {Array<string>} record.tags
+     * @param {string} record.attachmentKey
+     * @param {string} record.customDate The custom date entered by user
+     * @param {string} record.createdAt The date of creation of record
+     * @returns {Promise<Object>}
+     */
+    updateFhirRecord(ownerId, record) {
```
Harsimran1: This probably needs recordId as second parameter.

Glumli: And the record as third?

Harsimran1: yes. Because update normally means update against `id`. But our documentService paramters differs in that sense. So maybe not such a good idea.

Harsimran1:
@Glumli @ibihim PTAL

src/lib/models/HCDocument.js
```diff
@@ -6,7 +6,7 @@ class HCDocument {
         files = [],
         type = 'Document',
         title,
-        creationDate = new Date(),
+        customDate,
```
Harsimran1: Will it make sense to set default customDate?

Glumli: Is it required in FHIR?

Harsimran1: it is required for sol. But we're by default setting new Date() in fhirservice.

src/services/fhirService.js
ibihim: Isn't the id within the record?

src/services/fhirService.js
```diff
@@ -8,46 +8,85 @@ import createCryptoService from './cryptoService';
...
+     * @returns {Promise<Object>}
+     */
+    uploadFhirRecord(ownerId, record, uploadRequest) {
+        return fhirValidator.validate(record.fhirResource)
             .then(() => {
-                const tags = [
-                    ...customTags,
-                    ...taggingUtils.generateTags(fhirResource),
+                record.tags = [
+                    ...(record.tags || []),
```
ibihim: good catch

src/services/fhirService.js
```diff
@@ -8,46 +8,85 @@ import createCryptoService from './cryptoService';
...
+    uploadRecord(ownerId, record, uploadRequest) {
         let owner;
         return userService.getUser(ownerId)
             .then((user) => {
                 owner = user;
                 return Promise.all([
-                    createCryptoService(owner.id).encryptObject(resource),
-                    Promise.all(tags.map(tag => hcCrypto.symEncryptString(owner.tek, tag))),
+                    createCryptoService(owner.id).encryptObject(record.fhirResource),
+                    Promise.all(record.tags.map(tag => hcCrypto.symEncryptString(owner.tek, tag))),
```
ibihim: Why do we have a Promise.all in a Promise.all?

src/lib/models/HCDocument.js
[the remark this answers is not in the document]
Harsimran1: creationDate is different than customDate(It is something user can enter about when that document was actually created- human inputted date). creationDate is machine generated date which is part of our models, but not exposed to SDK developers to be edited.

src/services/documentService.js
```diff
@@ -68,10 +68,10 @@ const documentService = {
         if (!hcDocumentUtils.isValid(hcDocument)) {
             return Promise.reject(new ValidationError('Not a valid hcDocument'));
         }
-
+        hcDocument.creationDate = hcDocument.creationDate || new Date();
```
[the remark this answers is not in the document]
Harsimran1: This should probably be `hcDocument.creationDate = new Date();`

src/services/documentService.js
```diff
@@ -68,10 +68,10 @@ const documentService = {
...
         return this.fhirService.createFhirRecord(
             ownerId,
-            hcDocumentUtils.toFhirObject(hcDocument, this.clientId),
+            { fhirResource: hcDocumentUtils.toFhirObject(hcDocument, this.clientId) },
```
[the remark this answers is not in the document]
Harsimran1: It was refactoring involved to be able to do that ticket. Otherwise the function may had do many parameters.

src/services/fhirService.js
Glumli: Cause when having Promise.all([Promise.resolve, [unresolved Promises]]) it resolves, as far as I know

src/lib/models/utils/hcDocumentUtils.js
```diff
@@ -73,19 +82,19 @@ const hcDocumentUtils = {
             resourceType: 'DocumentReference',
             status: 'current',
             type: { text: hcDocument.type },
-            indexed: hcDocument.creationDate.toISOString(),
             author: [{ reference: '#contained-author-id' }],
             description: hcDocument.title,
             subject: { reference: hcDocument.title },
             contained: [],
         };
+        if (hcDocument.creationDate) fhirObject.indexed = hcDocument.creationDate.toISOString();
```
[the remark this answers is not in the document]
Harsimran1: i think the current eslint configurations allows. It was added at other places because work limit. But I always find them prettier. But will change if you insist.

src/lib/models/HCDocument.js
Harsimran1: tbh there isn't much we can do about naming at this point because this name been implemented in other SDKs one month back. (it was not implemented in JS for some reason. ) And `creationDate` is something else already which is by default set by the system and is one of lifetime requirements. Even though it is not set by developer, I can be returned while downloading documents.

src/lib/models/HCDocument.js
```diff
@@ -17,13 +17,13 @@ class HCDocument {
                 && typeof type === 'string'
                 && (!annotations || typeof annotations === 'object')
                 && typeof title === 'string'
-                && creationDate instanceof Date
+                && (!customDate || customDate instanceof Date)
```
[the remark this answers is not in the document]
Harsimran1: afait !customDate can be true in case of `null, undefined, "", true ` and `customDate instanceof Date` is `false` in all three cases

Harsimran1:
@ibihim @Glumli PTAL

Glumli:
Did you test using the example app already?

src/lib/models/HCDocument.js
```diff
@@ -6,7 +6,7 @@ class HCDocument {
         files = [],
         type = 'Document',
         title,
-        creationDate = new Date(),
+        customDate = new Date(),
```
Glumli: Could you please validate this naming with Peter/Tobi Im not a fan of "customDate" at all. Then we can also call it "customTitle" and "customType"

src/lib/models/HCDocument.js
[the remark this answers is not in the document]
ibihim: oh, I think I got it wrong. Why are we throwing an error, if customDate is of type Date?

src/services/fhirService.js
ibihim: wait what? let me double check that...

ibihim: Its not true, Promise.all only resolves, when the Promises left the pending state.
Try that:
```JavaScript
const rnd100 = () => Math.floor(Math.random() * 100);
const makeRNDPromsie = tm => new Promise((resolve, reject) =>
  setTimeout(() => resolve(tm), tm)
);

Promise.all([makePromise(rnd100()), makePromise(rnd100())]).then(console.log);
```
As you will see, the response will contain something like `[7, 51]` (afer 51 seconds) and not `[7, Promise{<pending>, ...}]`

ibihim: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all

> 

The Promise.all(iterable) method returns a single Promise that resolves when all of the promises in the iterable argument have resolved or when the iterable argument contains no promises. It rejects with the reason of the first promise that rejects.

Glumli: And now try this:
```const rnd100 = () => Math.floor(Math.random() * 100);
const makeRNDPromsie = tm => new Promise((resolve, reject) =>
  setTimeout(() => resolve(tm), tm)
);

Promise.all([[makePromise(rnd100()), makePromise(rnd100())]).then(console.log);```

File: test/testUtils/documentResources.js [hunk 12]
```diff
@@ -3,6 +3,8 @@ import hcSpecialty from '../../src/lib/models/HCSpecialty';
 const documentResources = {
     files: [new File(['test'], 'testName'), new File(['test'], 'testName')],
     creationDate: new Date('Thu, 23 Nov 2017 22:57:55 GMT'),
+    customDate: new Date('December 17, 1995'),
+    updatedDate: new Date('2018-12-17T03:24:00'),
```
ibihim: great!

src/services/fhirService.js
ibihim: oh, what a nasty construct

ibihim: One could do (not a change request, *ok* with double Promise.all):
```JavaScript
Promise.all(
     ...createCryptoService(owner.id).encryptObject(record.fhirResource),
     ...record.tags.map(tag => hcCrypto.symEncryptString(owner.tek, tag)),
).then(([data, key, ...encryptedTags]) => {
    // ...
})
```
But not really wanting to try that out, as I am worried about all the spreading forth and back (performance).

Harsimran1:
It is tested with terra with nothing to change in terra. So i guess example app should work too. Example apps setup is too complicated to make test

src/lib/models/HCDocument.js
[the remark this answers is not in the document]
Harsimran1: We are not throwing error if `customDate instanceof Date`. There is a `!` operator in the beginning. We need to check falsify of customDate because it is not a required parameter. We dont want to throw error in case of undefined. e.g. in case of undefined customDate we need `!(!customDate || customDate instance of Date)` to be false,  in case of valid date format, it should be false and in case of invalid it should be true and throw an error

PR #173, opened 2018-06-18
refactors update document

Glumli:


src/services/documentService.js
```diff
@@ -91,58 +104,44 @@ const documentService = {
...
-            documentRoutes.downloadRecord(ownerId, hcDocument.id);
-
-        const encryptedFilesPromise = recordPromise
-            .then(record => createCryptoService(ownerId).encryptBlobs(
-                newAttachments.map(attachment => attachment.file),
-                record.attachment_key,
-            ));
+        const encryptedFilesPromise = attachmentKeyPromise
+            .then(ak =>
+                createCryptoService(ownerId).encryptBlobs(
```
ibihim: I don't know if you like challenges or at how much time pressure you are, but this could be a nice memoization application :smile: 
Reason: spinning up a cryptoService on every updateDocument call might be expensive...

src/services/documentService.js
```diff
@@ -91,58 +104,44 @@ const documentService = {
...
-                    }));
-                return uploadFilePromise;
-            })
-            .then(() => {
-                const customTags = hcDocument.annotations
-                    ? taggingUtils.generateCustomTags(hcDocument.annotations)
-                    : [];
+            .then(([[encryptedFiles, attachmentKey], uploadInformation]) => {
+                const uploadFilesPromise = encryptedFiles.map((encryptedFile, index) => {
+                    newAttachments[index].id = uploadInformation[index].id;
```
ibihim: Can we guarantee order (uploadInformation <=> encryptedFiles)? Otherwise it will cause some very hard to reason issues.

src/services/documentService.js
```diff
@@ -7,6 +7,19 @@ import ValidationError from '../lib/errors/ValidationError';
...
+    const newAttachments = [];
+    const oldAttachments = [];
+    collection.forEach((element) => {
+        if (element[attribute]) {
+            oldAttachments.push(element);
+        } else {
+            newAttachments.push(element);
+        }
+    });
+    return { oldAttachments, newAttachments };
```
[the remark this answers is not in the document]
ibihim: Hint: When trying to solve a problem, try to abtract what you are doing. If you would do so, you could see, that you are solving here a general problem and not a particular one :smile:

src/services/documentService.js
```diff
@@ -91,58 +104,44 @@ const documentService = {
...
                     ownerId,
                     hcDocument.id,
                     hcDocumentUtils.toFhirObject(hcDocument, this.clientId),
-                    customTags,
+                    taggingUtils.generateCustomTags(hcDocument.annotations),
                     attachmentKey,
                 );
+                return Promise.all([
+                    updateFhirRecordPromise,
+                    Promise.all(uploadFilesPromise),
```
ibihim: Is this some common pattern to have Promise.all wrapped in another Promise.all? why???

src/services/fhirService.js
```diff
@@ -6,25 +6,20 @@ import dateUtils from '../lib/dateUtils';
 import hcCrypto from '../lib/crypto';
 import createCryptoService from './cryptoService';
 
-
 const fhirService = {
-    updateFhirRecord(ownerId, recordId, fhirObject, customTags = [], attachmentKey = null) {
+    updateFhirRecord(ownerId, recordId, fhirResource, customTags = [], attachmentKey) {
```
ibihim: why did you remove attachmenKey = null?

src/services/documentService.js
```diff
@@ -91,58 +104,44 @@ const documentService = {
...
+            .then(ak =>
+                createCryptoService(ownerId).encryptBlobs(
+                    newAttachments.map(attachment => attachment.file),
+                    ak,
+                ));
 
         return Promise.all([
             encryptedFilesPromise,
             newAttachments.length ?
                 documentRoutes.getFileUploadUrls(ownerId, hcDocument.id, newAttachments.length) :
```
ibihim: getFileUploadURLs :smile:

src/services/documentService.js
ibihim: I see, it is not important :+1:

src/services/documentService.js
```diff
@@ -91,58 +104,44 @@ const documentService = {
...
+            .then(([[encryptedFiles, attachmentKey], uploadInformation]) => {
+                const uploadFilesPromise = encryptedFiles.map((encryptedFile, index) => {
+                    newAttachments[index].id = uploadInformation[index].id;
+                    return fileRoutes.uploadFile(
+                        uploadInformation[index].sas_token,
+                        encryptedFile);
+                });
                 hcDocument.attachments = [...oldAttachments, ...newAttachments];
-                return this.fhirService.updateFhirRecord(
+                const updateFhirRecordPromise = this.fhirService.updateFhirRecord(
```
ibihim: Why is it `this.fhirService` and not fhirSerivce???

ibihim: updateFHIRRecord....

src/services/documentService.js
Glumli: I answered this in Simranns PR already ;)

src/services/fhirService.js
Glumli: Does it make sense to have it? Is null so much better than undefined?

src/services/documentService.js
ibihim: And I re-answered to this in Simran's PR too

src/services/fhirService.js
ibihim: So, this indicates, that the value is optional. Just looks strange, that an optional value is followed by a non-optional one. 
I think this is common in TypeScript or so (declaring optionality by that).

src/services/documentService.js
ibihim: Ok, got it.

src/services/documentService.js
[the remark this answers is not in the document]
Glumli: Ah fuck... I was pulling my changes over to a new branch, as the other one had flow. 
But actualy I pulled the method out and also gave it a general name and the general parameter because I did exactly this. Forgot renaming the arrays to sth general after rewrite.

src/services/documentService.js
Glumli: We really have to think about how to do this in the future... I liked creating a new one each time, as the commonKey might change :/ And without memoization the propability for using an olc commonKey is smaller

src/routes/documentRoutes.js
```diff
@@ -5,11 +5,11 @@ const apiUrl = config.api;
 
 const documentRoutes = {
 
-    getFileDownloadUrl(ownerId, recordId, fileId) {
+    getFileDownloadURL(ownerId, recordId, fileId) {
         return hcRequest.submit('GET', `${apiUrl}/users/${ownerId}/documents/${recordId}/files/${fileId}/download_access_token`, { authorize: true, ownerId });
     },
 
-    getFileUploadUrls(ownerId, recordId, fileNumber) {
+    getFileUploadURLs(ownerId, recordId, fileNumber) {
```
ibihim: great

src/services/documentService.js
ibihim: ok

PR #174, opened 2018-06-18
adds flow types

Glumli:


Glumli:
@ibihim why does it suck?

Glumli:
BTW: Seems like Terra is also using Flow

Glumli:
@ibihim why does "this" suck?

ibihim:
let put it offline. open sauce talk D:

PR #175, opened 2018-06-20
Simplify isExpired check

Glumli:


src/lib/hcRequest.js
```diff
@@ -27,12 +27,24 @@ const hcRequest = {
         this.accessTokens = {};
     },
 
-    setMasterAccessToken(accessToken) {
-        this.masterAccessToken = `Bearer ${accessToken}`;
+    setAccessToken(userId, accessToken) {
+        this.accessTokens[userId] = {
+            access_token: accessToken.access_token,
+            expires: Date.now() + (accessToken.expires_in * 1000), // seconds to milliseconds
```
ibihim: I think that the accessTokens issued by Vega have an `exp` property, which is the System Current Millis, which would make it possible to create the Date like that: `new Date(accessToken.exp)`.
How does the accessToken obj look like? I am just curious, why you strip stuff out and recreate it.

src/lib/hcRequest.js
```diff
@@ -27,12 +27,24 @@ const hcRequest = {
...
+        };
     },
 
-    setAccessToken(userId, accessToken) {
-        this.accessTokens[userId] = `Bearer ${accessToken}`;
+    setMasterAccessToken(token) {
+        this.masterAccessToken = {
+            access_token: token.access_token,
+            expires: Date.now() + (token.expires_in * 1000), // seconds to milliseconds
```
ibihim: This is duplicate logic by now...

src/lib/hcRequest.js
```diff
@@ -27,12 +27,24 @@ const hcRequest = {
...
+        };
+    },
+
+    parseToken(accessToken) {
```
ibihim: This can be an external function. I wouldn't call it prarseToken. I would call it rather, bearerifier, createAuthorizationHeader or something like that.

src/lib/hcRequest.js
```diff
@@ -7,8 +7,8 @@ const maxRetries = 2;
 
 const isHealthCloudPath = path => path.startsWith(config.api);
 
-const isExpired = error =>
-    error.status === 401 && error.body && error.body.error && error.body.error.includes('expired');
+const isExpired = (accessToken = {}) =>
+    accessToken.expires < Date.now();
```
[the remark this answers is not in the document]
Glumli: Do I? To check if it is expired I don't need this, do I?

src/lib/hcRequest.js
Glumli: The one I gout back in Terra had expires_in

src/lib/hcRequest.js
Glumli: Can I fix this in another PR tomorrow?

src/lib/hcRequest.js
Glumli: Why should it be external?

src/lib/hcRequest.js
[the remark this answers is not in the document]
Harsimran1: but doesn't 401 already indicates expiry?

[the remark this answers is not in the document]
Glumli: Now we check if it is expired before sending the request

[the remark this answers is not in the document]
Glumli: Afterwards we are just looking for 401 and then getting new accessTokens

src/lib/hcRequest.js
Glumli: For the normal accessToken we get back expires_in: 3600.
I cahnged it for the sdk to calculate the time, so that the customer doesn't have to do it in their backend

src/lib/hcRequest.js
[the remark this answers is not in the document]
Harsimran1: hmm, i'm not sure. At any 401 we anyways refresh it, then what is the need to check the expired case before. I know expired_in comes into play if tokens are set as cookies (as their expiry can be detected when cookie expiries ), but not sure if this is the right way to go about it. Need to read more.  As we're making the code and initialisation very complicated here.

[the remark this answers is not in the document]
ibihim: I didn't say that yo need to check the type of accessToken.expires, I am saying, that you are sure that it is of type number?!
@Harsimran1 It is better to check if an access token is expired. So you know upfront and don't need to verify that by waiting for the 401 to come back. To make it clear, if you check the expiry, you have one call less every X minutes.

src/lib/hcRequest.js
```diff
@@ -27,12 +27,19 @@ const hcRequest = {
...
+        this.accessTokens[userId] = this.importToken(accessToken);
     },
 
-    setAccessToken(userId, accessToken) {
-        this.accessTokens[userId] = `Bearer ${accessToken}`;
+    setMasterAccessToken(accessToken = {}) {
+        this.masterAccessToken = this.importToken(accessToken);
+    },
+
+    importToken(accessToken) {
```
ibihim: I am still not seeing, why we are bloating up hcRequest with this. It is not releated to any this context. I would define that function outside of hcRequest, so this abomination does not get even bigger. And I am not a fan of mixing the bearer_token things with expiry.

src/lib/hcRequest.js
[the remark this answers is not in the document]
Glumli: I would suggest to create a seperate Ticket for this. The ticket I got actually doesn't require this change. I'll then just ignore the time in Terra for now and simply give back a token, when asked

PR #179, opened 2018-06-21
Js/hc 1556 add model version

Glumli:


src/services/fhirService.js
```diff
@@ -98,6 +98,7 @@ const fhirService = {
                     encrypted_tags: results[1],
                     date: dateUtils.formatDateYyyyMmDd(record.customCreationDate || new Date()),
                     attachment_key: record.attachmentKey,
+                    data_model_version: DATA_MODEL_VERSION,
```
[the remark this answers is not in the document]
Glumli: Uh wow. Your right. Just checked iOS

package.json
```diff
@@ -1,6 +1,9 @@
 {
   "name": "hc-sdk-js",
   "version": "0.1.0",
```
ibihim: We should start to use the version field within package.json

test/services/fhirServiceTest.js
```diff
@@ -139,6 +139,7 @@ describe('services/fhirService', () => {
         }).default;
 
         fhirServiceUploadRecordSpy = sinon.spy(fhirService, 'uploadRecord');
+        global.DATA_MODEL_VERSION = 1;
```
ibihim: Where do you verify, that the DATA_MODEL_VERSION matches what you set here?

Glumli:
@ibihim PTAL

PR #181, opened 2018-06-22
HC-1654: fixes deployment

Glumli:


Glumli:
@theintz @Harsimran1 @pussinboots PTAL
we will have to update the environment, but the dev credentials from 1PW failed

.travis.yml
```diff
@@ -11,66 +11,62 @@ before_install:
...
-        - npm run build:${ENV}
+        - npm run build
+        - sha256sum dest/healthcloud_sdk.js > dest/healthcloud_sdk.js.sha256
       deploy:
         provider: script
         script: scripts/publish_sdk.sh ${ENV}
         skip_cleanup: true
         on:
-          branch: master
+          branch: js_th/improves-deployment
```
Harsimran1: Make sure to change this to master

.travis.yml
```diff
@@ -11,66 +11,62 @@ before_install:
...
     - stage: build production
       node_js: "7"
       env:
-      - ENV=sandbox # change to PRODUCTION as soon as endpoint is available
+        - ENV=dev # change to PRODUCTION as soon as endpoint is available
```
Harsimran1: do we want to change this to `prod` now?

pussinboots:
The production endpoint is available so you can change that
- ENV=dev # change to PRODUCTION as soon as endpoint is available
to
- ENV=prod # change to PRODUCTION as soon as endpoint is available.

.travis.yml
Glumli: Good catch

Glumli:
@pussinboots You're totally right. Can you please make this change, as I don't have access to the azure credentials for prod?

PR #184, opened 2018-07-12
H/add typescript

Glumli:


src/lib/models/utils/hcDocumentUtils.ts
```diff
@@ -1,8 +1,9 @@
-import Ajv from 'ajv';
+const Ajv = require('ajv');
```
Harsimran1: why this?

Glumli: otherwise it doesn't compile. 
I will take a look into it once again, but I was sitting on it for 30 mins or so

src/lib/models/utils/hcDocumentUtils.ts
```diff
@@ -90,35 +92,34 @@ const hcDocumentUtils = {
...
                     hcDocument.author.specialty,
                 ),
-            };
-        }
+            } :
+            null;
 
-        if (hcDocument.additionalIds) {
-            fhirObject.identifier = Object.keys(hcDocument.additionalIds).map(key => ({
+        const identifier = (hcDocument.additionalIds) ?
```
Harsimran1: fhir `identifier` should be an array. Setting `null` will fail the fhir validation. This results in upload failure. Also it would be better to not set `null`, `undefined`, or `empty arrays` at all as they are not required fields.

src/lib/models/utils/hcDocumentUtils.ts
```diff
@@ -90,35 +92,34 @@ const hcDocumentUtils = {
...
+            Object.keys(hcDocument.additionalIds).map(key => ({
                 value: hcDocument.additionalIds[key],
                 assigner: {
                     reference: key,
                 },
-            }));
-        }
+            })) :
+            null;
```
Harsimran1: did u test this with terra. I'm guesssing its supposed to be array

src/lib/models/utils/hcDocumentUtils.ts
Harsimran1: can u looks into this? if taking longer, i can do it too

tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "outDir": "tsDist",
+    "module": "es6",
+    "target": "es6",
+    "allowSyntheticDefaultImports": true
+  },
+  "exclude": [
+    //"node_modules",
```
Harsimran1: remove this line ? or do we want to add node_modules here?

